PR #74, opened 2022-12-19
[FEAT] Add step size argument to cross validation method

FedericoGarza:
## Description

This PR adds the `step_size` argument to the cross validation method. 
The argument controls the size between each cross validation window.

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.


jmoralez:
@FedericoGarza WDYT about making this an optional int in the forecast objects as well to not make it a breaking change?

jmoralez:
Hmm I see that's the default in statsforecast. I think we can maybe print a warning instead.

FedericoGarza:
hey @jmoralez! Thanks for your feedback! I've added a warning message explaining the new behavior [here](https://github.com/Nixtla/mlforecast/pull/74/commits/73b132d08509eedc6d38b332e80b43de5945c7e6). WDYT?

mlforecast/forecast.py
```diff
@@ -299,6 +299,15 @@ def cross_validation(
         result : pandas DataFrame
             Predictions for each window with the series id, timestamp, last train date, target value and predictions from each model.
         """
+        if window_size != step_size:
+            warning_msg = (
+                "The gap between each cross validation window (controled by `step_size`) "
+                "changed from being equal to the `window_size` to being equal to 1. "
```
mergenthaler, line 305: This error message could be a little bit clearer.

PR #79, opened 2022-12-23
[FEAT] Add `new_data` argument to `predict` method (allow transferability)

FedericoGarza:
## Description

This PR adds the `new_data` argument to the `predict` method.
The new argument is useful to train a model with some data and then use the fitted model to predict new data.

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.


mlforecast/core.py
```diff
@@ -245,10 +245,17 @@ def __init__(
     ):
         if isinstance(freq, str):
             self.freq = pd.tseries.frequencies.to_offset(freq)
+        elif isinstance(freq, pd._libs.tslibs.offsets.BaseOffset):
```
jmoralez, line 248: in case they move this to a different module.
```suggestion
        elif isinstance(freq, pd.offsets.BaseOffset):
```

mlforecast/core.py
```diff
@@ -245,10 +245,17 @@ def __init__(
...
+            self.freq = freq
```
jmoralez, line 249: we should document that `freq` can also be a pandas offset by adding it to the `Freq` type above and to the docstrings.

mlforecast/core.py
```diff
@@ -245,10 +245,17 @@ def __init__(
...
         elif isinstance(freq, int):
             self.freq = freq
         elif freq is None:
             self.freq = 1
+        else:
+            raise Exception(
```
jmoralez, line 255:
```suggestion
            raise ValueError(
```

mlforecast/distributed/forecast.py
```diff
@@ -207,6 +207,7 @@ def predict(
         dynamic_dfs: Optional[List[pd.DataFrame]] = None,
         before_predict_callback: Optional[Callable] = None,
         after_predict_callback: Optional[Callable] = None,
+        new_data: Optional[pd.DataFrame] = None,
```
jmoralez, line 210: This should probably be dask dataframe. But maybe it isn't worth having this in the distributed interface. I guess it's more common to do distributed training and then save the trained model for predicting smaller datasets locally, we should probably have a `to_local` method instead so that after training users can convert the distributed forecast to a local one and then use that one with `new_data`. WDYT?

mlforecast/forecast.py
```diff
@@ -233,7 +238,30 @@ def predict(
...
+        if new_data is not None:
+            new_ts = TimeSeries(
+                self.ts.freq,
+                self.ts.lags,
+                self.ts.lag_transforms,
+                self.ts.date_features,
+                self.ts.differences,
+                self.ts.num_threads,
+            )
+            new_ts.fit_transform(
```
jmoralez, line 251: The transform part isn't necessary. We can probably move the checks that are done inside fit_transform before calling _fit into _fit, if that's what you're worried about.
```suggestion
            new_ts._fit(
```

nbs/core.ipynb
```diff
@@ -1762,7 +1784,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
```
jmoralez, line 1787: I have the action_files/clean_nbs script to avoid exactly this diff.

jmoralez (changes requested):

mlforecast/core.py
FedericoGarza: fixed

mlforecast/core.py
FedericoGarza: fixed

mlforecast/core.py
FedericoGarza: fixed

mlforecast/forecast.py
FedericoGarza: fixed

nbs/core.ipynb
FedericoGarza: fixed

mlforecast/distributed/forecast.py
FedericoGarza: I agree! Maybe we can leave the possibility to use the new feature in a distributed way (I already corrected the typing) and open an issue to include the `to_local` method. WDYT?

mlforecast/distributed/forecast.py
```diff
@@ -224,13 +225,50 @@ def predict(
             Function to call on the predictions before updating the targets.
                 This function will take a pandas Series with the predictions and should return another one with the same structure.
                 The series identifier is on the index.
+        new_data : pandas DataFrame, optional (default=None)
```
jmoralez, line 228:
```suggestion
        new_data : dask DataFrame, optional (default=None)
```

jmoralez (changes requested):

PR #86, opened 2023-02-01
Add prediction (conformal) intervals

FedericoGarza:
## Description

This PR adds conformal prediction intervals to MLForecast.
The implementation is based on the paper [Conformal Time Series Forecasting](https://proceedings.neurips.cc/paper/2021/hash/312f1ba2a72318edaaa995a67835fad5-Abstract.html).

Checklist:
- [x] This PR has a meaningful title and a clear description.
- [x] The tests pass.
- [x] All linting tasks pass.
- [x] The notebooks are clean.

PR #87, opened 2023-02-02
Add prediction intervals tutorial

FedericoGarza:
## Description

This PR adds a prediction intervals tutorial.
It depends on #86.

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.


jmoralez:
The failing test is flaky so you can safely ignore it. Are the notebooks that big because of the plots?

FedericoGarza:
> The failing test is flaky so you can safely ignore it. Are the notebooks that big because of the plots?

Yes, I think so. The plots use `plotly` and they are interactive. xD Would you recommend using `matplotlib` instead?

jmoralez:
It's fine. I changed the other plots to saving them instead to avoid diffs because of the plot sizes (for some reason different python versions were generating different grid sizes even though we set the figsize), but hopefully that won't happen with these and they probably won't be modified as often as the other ones.

jmoralez (approved):

PR #101, opened 2023-03-07
vectorize prediction intervals creation

jmoralez:
## Description

Uses the fact that the predictions from CV are sorted by serie and window to compute the intervals for all series at once. Moves fugue to be a dependency for distributed.

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.


FedericoGarza (approved):
nice! LGTM :)

PR #104, opened 2023-03-20
add ray integration

FedericoGarza:
## Description

This PR integrates ray to the `DistributedMLForecast` class.
This new integration allows training lighgbm and xgboost models using ray dataframes.

Checklist:
- [ ] The tests pass.
- [x] All linting tasks pass.


jmoralez:
The [Scikit-Learn](https://docs.ray.io/en/latest/ray-air/trainer.html#scikit-learn-trainer) is not distributed. 
image
We need to incorporate the [LightGBMTrainer](https://docs.ray.io/en/latest/ray-air/trainer.html#lightgbmtrainer) and the [XGBoostTrainer](https://docs.ray.io/en/latest/ray-air/trainer.html#xgboost-trainer) or directly use lightgbm_ray and xgboost_ray

FedericoGarza:
hey @jmoralez! Yes, I'm working on including the distributed trainers. I think it is a good option to keep the scikit-learn trainer since the training part for those models uses just one actor of the cluster, which can be slow in some cases or throw memory errors. Still, the users will have the option if they want to do it (I tested it on anyscale and it works as expected). WDYT?

jmoralez:
What's the difference vs the regular MLForecast?

FedericoGarza:
I would say that it allows training on ray datasets distributively (using the distributed trainers) but also classical sklearn models without changing between pandas and ray datasets.

environment.yml
```diff
@@ -20,4 +20,7 @@ dependencies:
   - xgboost
   - pip:
     - datasetsforecast
+    - fugue[ray]
```
jmoralez, line 23: There's already a fugue in the conda section, can you remove that one?

mlforecast/distributed/forecast.py
```diff
@@ -208,6 +217,8 @@ def _preprocess_partitions(
             schema="ts:binary,train:binary,valid:binary",
             engine=self.engine,
             as_fugue=True,
+            partition=id_col,
```
jmoralez, line 220: I don't support setting this unless we also set `num`, which should be a multiple of the number of machines. The reason is that we'll have as many tasks as series, which are probably over tens of thousands, which will introduce a lot of overhead and will use more memory since we'll have tens of thousands of objects being serialized.
Dask partitions based on the index and spark provides repartitionByRange, but I don't see something similar for ray.

mlforecast/distributed/forecast.py
```diff
@@ -325,8 +340,18 @@ def _fit(
             for name, model in self.models.items():
                 trained_model = clone(model).fit(X, y)
                 self.models_[name] = trained_model.model_
+        elif RAY_INSTALLED and isinstance(data, RayDataset):
+            for name, model in self.models.items():
+                X = RayDMatrix(
+                    prep.select_columns(cols=features + [target_col]),
+                    label=target_col,
+                )
```
jmoralez, line 348: This can be moved out of the for loop.

jmoralez (changes requested):

mlforecast/distributed/forecast.py
FedericoGarza: Yes, I found the problem with Ray where using only `repartition(num_partitions)` produces strange results (breaks certain series). I researched a bit and couldn't find a native way to create partitions based on a column and ended up hardcoding the `by` argument. Using fugue, I integrated the `num_partitions` argument in the newest commit to creating partitions based on `id_col`. If this argument is None, the original partitions of Dask and Spark will be respected; in that case, the partitions will be done by `id_col` for Ray. If `num_partitions` is used, the partitions will be overwritten. I included several tests to ensure that `partition_results` object contains the expected number of rows and that the number of partitions is correct for Spark, Dask, and Ray. I also included a test to show that the same metadata of the dataframes is returned before and after using `num_partitions`. With this change, for example, in Dask, it is no longer necessary to partition the dataframe by `id_col`. Let me know what you think :)

mlforecast/distributed/forecast.py
FedericoGarza: Thanks for catching that. Fixed in the newest commit.🙌

environment.yml
FedericoGarza: Yes, sure. `fugue[ray]` install other libraries for Ray and fugue to run properly. I tried integrating them in the conda section but some wheels are unavailable for py3.7. Should I include them separately in the pip section?

jmoralez: I meant remove the ray from the conda section and leave this one

FedericoGarza: hahaha I see, sorry. Fixed.

mlforecast/distributed/forecast.py
```diff
@@ -191,6 +205,19 @@ def _preprocess_partitions(
         window_info: Optional[WindowInfo] = None,
         fit_ts_only: bool = False,
     ) -> List[Any]:
+        if self.num_partitions:
+            partition = dict(by=id_col, num=self.num_partitions, algo="coarse")
```
jmoralez, line 209: I couldn't find what the `algo='coarse'` does. I think the default (hash) is probably fine. Is there a reference you can share to take a look?

mlforecast/distributed/forecast.py
jmoralez: I think that's a great approach! Just left a minor comment on the algo option.

mlforecast/distributed/forecast.py
FedericoGarza: That was a little bit tricky because I started playing around with just `dict(by=id_col, num=self.num_partitions)` and I found unexpected (for me) behavior. For example, in the following test, I expected to end up with just only two rows but I got three rows instead (same for spark):

image

After discussing it with the fugue guys, they suggested using `algo='coarse'` and it worked,

image

Basically, `num=2` (without `algo='coarse'`)  creates two partitions based on `id`  but inside those partitions, `get_n_ids` is applied to each of the ids.

jmoralez: Got it, that seems better indeed. Thanks for the explanation!

jmoralez (approved):
LGTM. We can probably ignore that failing test, seems to be related to the synapseml download

PR #504, opened 2025-07-11
add future X_df features to predict

JLiekenbrock:
## Description
In this PR MLForecast Core is updated so that using predict with direct approach will correctly use the future external regressors.

Checklist:
- [ ] The tests pass.
- [ ] All linting tasks pass.


JLiekenbrock:
I started having a look at the fit part, I'll try to add it to.

PR #506, opened 2025-07-23
add the possibility to pass additional arguments in the fit function

PierD86:
## Description
This PR adds support for passing model-specific arguments to the MLForecast.fit() method. This enhancement allows users to directly supply parameters such as eval_set and early_stopping_rounds to underlying regressors like LGBMRegressor or CatBoostRegressor.

This change removes the need to rely on LightGBMCV, enabling broader compatibility with early stopping, a widely-used regularization technique that does not depend on cross-validation. Furthermore, each regressor can have crucial parameters to be passed during fit.

Moreover, the current workaround using custom training pipelines (as outlined in the documentation) adds unnecessary complexity. This update simplifies the workflow and improves usability for MLForecast users.

Note: I could not check the new code using notebooks, also, the original version of forecast.ipynb is not working locally on my PC.
The error, also running pre-commits is the following: 
`ImportError: Could not import '__path__' from fastcore.dispatch - this module has been moved to the fasttransform package.`

However, the code is simple and easy to check.

Checklist:
- [x ] This PR has a meaningful title and a clear description.
- [ ] The tests pass. 
- [ ] All linting tasks pass.
- [x ] The notebooks are clean.


nasaul (approved):
Sorry for the delayed review @PierD86. We value and are grateful for your contribution. I'll work on improving the reviewing times for contributions. Again sorry for the delay.

Issue #358, opened 2024-06-20
[core] perform more validations on update


jmoralez:
### Description

We currently don't check that the dates in the dataframe provided to the update method are actually in the future, so users could append already seen values, we should check that.

### Use case

Allow users to more confidently call the update method and know that if it didn't raise any errors everything went well.

PR #541, opened 2025-12-14
Enhancement/validations on update

janrth:
Tries to solve #358

Validates that the update df has the expected shape so that each unique_id starts from the last ds as seen in the previous df and contains the expected number of ds.

For each unique_id the number of ds date points are counted from the observed update df and this is then compared to the expected number of date points, which is calculated by the estimated start and end date. The estimated start date is observed based on the stored series last date + offset(freq).

There is an option to turn off the validate_input step. While overall the performance is pretty fast, it might be a bit annoying if one has hundreds of millions of rows. 

Initially I started just checking if the first ds of the update is in the future for each unique_id, but then I felt this is not checking much really and started to implement a stronger logic. The issue itself is a bit vague and I am open for any changes as I implemented based on my interpretation of the task.

## Description
Tries to implement more checks on update df

Checklist:
- [ ] The tests pass.
- [x] All linting tasks pass.

### DIFF
--- a/mlforecast/core.py
+++ b/mlforecast/core.py
@@ -865,6 +865,7 @@ class TimeSeries:
     
     def _validate_new_df(self, df: DataFrame) -> None:
         if isinstance(df, pl.DataFrame):
+            df = df.sort([self.id_col, self.time_col])
             stats = (
                 df.group_by(self.id_col)
                 .agg(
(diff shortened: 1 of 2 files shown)


nasaul (commented):
Overall looks good, however in order to merge you should address the following:

- Polars categorical encoding mismatch ([core.py](https://github.com/janrth/mlforecast/blob/f80651d3ae06c15fe139123def29e344323aa8f8/mlforecast/core.py#L884)) - All 5 Polars tests fail. The join operation fails because the categorical columns have different encodings. Needs string casting before join, like the pandas branch does.
- Type hint is wrong ([core.py](https://github.com/janrth/mlforecast/blob/f80651d3ae06c15fe139123def29e344323aa8f8/mlforecast/core.py#L866)) - Says pd.DataFrame but should be DataFrame since it handles both pandas and polars
- Add docstring for `validate_input` parameter in both both [forecast.py](https://github.com/janrth/mlforecast/blob/f80651d3ae06c15fe139123def29e344323aa8f8/mlforecast/forecast.py#L964) and [core.py](https://github.com/janrth/mlforecast/blob/f80651d3ae06c15fe139123def29e344323aa8f8/mlforecast/core.py#L974)

nasaul (commented):
Hey Jan, I've actually updated the tests in order to capture different frequencies and the solution doesn't hold. We have to use the offset from utilsforecast `ufp.offset_times` in order to make it work.

Issue #529, opened 2025-11-27
[MLForecast: Distributed] `AttributeError: MLForecast has no attribute _cs_df`


yherin:
## Actual behaviour

1. Fit a forecasting model using DistributedMLForecast, do not specify intervals.
2. Call `to_local()` to get a MLForecast object
3. Call `save()` on the mlforecast object
4. `AttributeError: `MLForecast has no attribute _cs_df`

## Expected behaviour

No attribute error when calling `.save()`

## Workaround

```
mlf = dmlf.to_local()
mlf._cs_df = None
mlf.save(outpath)
```

## Proposed fix

Hard to say where the bug comes from:
- Is it simply an oversight in the `save` method? -> add a safety check
- Or does `.to_local()` fail to set the `._cs_df` attribute as it should -> set _cs_df at the end of `to_local()`

### Versions / Dependencies

`mlforecast==1.0.2`

### Reproduction script

<no repo script>

### Issue Severity

Medium: It is a significant difficulty but I can work around it.

Issue #528, opened 2025-11-27
[Distributed|Core]  Calling `DistributedMLForecast.to_local()` with `Combine` transforms fails due to AttributeError


yherin:
## Bug description

The method `stack` assumes that the first item of `transforms` has the `_core_tfm` attribute. However, this is not the case for `Combine`, due to it having a different constructor compared to other transforms.

Compare 
```python
class Combine(_BaseLagTransform):
    def __init__(
        self, tfm1: _BaseLagTransform, tfm2: _BaseLagTransform, operator: Callable
    ):
        self.tfm1 = tfm1
        self.tfm2 = tfm2
        self.operator = operator
```

to

```python
class Lag(_BaseLagTransform):

    def __init__(self, lag: int):
        self.lag = lag
        self._core_tfm = core_tfms.Lag(lag=lag)
```

## Expected behaviour

All supported transforms can be properly serialised. 

## Stack trace

  ```python
   File "/app/src/workflows/forecasting.py", line 787, in fn_name
        mlf = dmlf.to_local()
              ^^^^^^^^^^^^^^^
      File "/opt/pysetup/.venv/lib/python3.12/site-packages/mlforecast/distributed/forecast.py", line 847, in to_local
        combined_core_lag_tfms = combine_core_lag_tfms(
                                 ^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/pysetup/.venv/lib/python3.12/site-packages/mlforecast/distributed/forecast.py", line 838, in combine_core_lag_tfms
        out[name] = partition_tfms[0].stack(partition_tfms)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/pysetup/.venv/lib/python3.12/site-packages/mlforecast/lag_transforms.py", line 66, in stack
        out._core_tfm = transforms[0]._core_tfm.stack(
                        ^^^^^^^^^^^^^^^^^^^^^^^
    AttributeError: 'Combine' object has no attribute '_core_tfm'. Did you mean: '_set_core_tfm'?
```

### Versions / Dependencies

mlforecast 1.0.2
coreforecast 0.0.16

### Reproduction script

```python
dmlf = DistributedMLForecast(
        models=[
            RayLGBMForecast()
        ],
        freq="D",
        lag_transforms=  { 1: [ Combine(
                RollingMean(window_size=7, min_samples=1),
                ExpandingMean(),
                operator.sub,
                )]},
       ...
)

[...]

PR #544, opened 2025-12-22
Fix issues with Ray and updates test suite

nasaul:
There were issues with the new Ray release, for this solution we need to update the requirements to the latest fugue implementation:

- [x] Update Fugue
- [x] Solves issues with combine in distributed environments #528
- [x] Solves issues with saving and add a test for that #529
- [x] Solves issue with fitting Ray datasets.
- [x] Adds new testing suite for Ray.

### DIFF
--- a/tests/distributed_ray/test_ray_forecast.py
+++ b/tests/distributed_ray/test_ray_forecast.py
@@ -15,7 +15,7 @@ warnings.simplefilter("ignore", FutureWarning)
 
 
 @pytest.mark.ray
-@pytest.mark.skipif(sys.version_info <= (3, 9), reason="Distributed tests are not supported on Python < 3.10")
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="Distributed tests are not supported on Python < 3.10")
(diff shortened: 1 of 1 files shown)


tests/distributed_ray/test_ray_forecast.py
```diff
@@ -0,0 +1,61 @@
...
+        },
+        date_features=["dayofweek"],
+    )
+    # df with features
+    training_df_featured = fcst.preprocess(
+        ray_dataset, static_features=[], dropna=False
+    )
+    fcst.fit(training_df_featured, static_features=[], dropna=False)
+    preds1 = fcst.predict(10).to_pandas().sort_values(["unique_id", "ds"], ignore_index = True)
+    fcst.save(f'/tmp/test_ray_forecast_model_{model_class.__name__}.pkl')
```
JQGoh: nit: In the future for test that requires saving to disk space, consider `tempfile` library so that we cleanup this automatically https://realpython.com/ref/stdlib/tempfile/

Issue #538, opened 2025-12-13
Tuning slow for huge data sets because backtest splits are created for each round


janrth:
### Description

For data sets with huge amounts of data, the tuning can be slow, because the backtest is created for each tuning round (as far as I see it) in _mlforecast_objective_.

Th core fitting methods are often fast because they are based on fast implementations from lgbm or xgb for example. But the backtest split can be a bottleneck.

I propose to leave the backtest split function as it is, but to re-use created splits (somehow) so the splits don't have to be created multiple times (as suggested by @yherin) 

### Use case

Faster tuning process due to reduction of multiple calculations of backtest splits.

PR #545, opened 2025-12-31
option to re-use cv splits during tuning

janrth:
At the moment the CV splits are created newly in each tuning trial, which can be slow for large datasets.
This PR adds an option to reuse CV splits across tuning trials.

When enabled, the splits are computed once in auto.py and passed into the optimization objective, where they are reused for all tuning trials. The default behavior is unchanged: if the option is disabled, CV splits are still generated inside each trial as before.

The original behavior is intentionally kept because this introduces a RAM vs CPU trade-off. Reusing splits can significantly reduce runtime when running many trials, but it keeps all train/validation splits in memory for the duration of the tuning run, which may increase peak memory usage on very large datasets.

To ensure correctness, a test was added using a deterministic model, verifying that predictions are identical when reusing CV splits versus recomputing them each trial.

When tested on the example below I saw a speed-up of 1.14x when re-using CV compared to the current implementation:

```
from utilsforecast.data import generate_series

series = generate_series(n_series=20000, freq='D', min_length=722, max_length=725, equal_ends=True)


# Setting up model
def my_lgb_config(trial: optuna.Trial):
    return {
        'learning_rate': 0.05,
        'verbosity': -1,
        'num_leaves': trial.suggest_int('num_leaves', 127, 128, log=True),
    }
    
def my_fit_config(trial: optuna.Trial):
    return {'dropna':False
    }
    
def my_init_config(trial: optuna.Trial):
    return {
        'lags': [24 * i for i in range(1, 3)],  # this won't be tuned
        'lag_transforms' : {
    1: [ExponentiallyWeightedMean(alpha=0.3),
        RollingMean(window_size=7, min_samples=1),
         RollingMean(window_size=21, min_samples=1),
          RollingMean(window_size=28, min_samples=1),
           RollingMean(window_size=50, min_samples=1),
        ],
}}

my_lgb = AutoModel(
    model=lgb.LGBMRegressor(),
    config=my_lgb_config,
)

import time

# -------------------------------
# Fitting model WITHOUT reusing CV splits
# -------------------------------
t0 = time.perf_counter()

auto_mlf_no_reuse = AutoMLForecast(
    models={"my_lgb": my_lgb},
    freq="D",
    season_length=24,
    fit_config=my_fit_config,
    init_config=my_init_config,
    reuse_cv_splits=False,
).fit(
    series,
    n_windows=15,
    h=horizon,
    num_samples=10,
)

t_no_reuse = time.perf_counter() - t0
print(f"Fit time without CV split reuse: {t_no_reuse:.2f} seconds")

# -------------------------------
# Fitting model WITH reusing CV splits
# -------------------------------
t0 = time.perf_counter()

auto_mlf_reuse = AutoMLForecast(
    models={"my_lgb": my_lgb},
    freq="D",
    season_length=24,
    fit_config=my_fit_config,
    init_config=my_init_config,
    reuse_cv_splits=True,   # <-- important
).fit(
    series,
    n_windows=15,
    h=horizon,
    num_samples=10,
)

t_reuse = time.perf_counter() - t0
print(f"Fit time with CV split reuse: {t_reuse:.2f} seconds")

# -------------------------------
# Comparison
# -------------------------------
speedup = t_no_reuse / t_reuse
delta = t_no_reuse - t_reuse

print(f"Speedup: {speedup:.2f}×")
print(f"Time saved: {delta:.2f} seconds ({delta / t_no_reuse * 100:.1f}%)")

```

## Description
Solves #538 

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -30,12 +30,12 @@ jobs:
       - name: Install dependencies (Linux/macOS)
         if: matrix.os != 'windows-latest'
         run: |
-          pip install uv pytest && uv pip install --system ".[dev]" fugue[dask,spark]>=0.8.1 "dask<=2024.12.1"
+          pip install uv pytest && uv pip install --system ".[dev,dask,ray]"
 
       - name: Install dependencies (Windows)
         if: matrix.os == 'windows-latest'
         run: |
-          pip install uv pytest && uv pip install --system ".[dev]" fugue[dask]>=0.8.1 "dask<=2024.12.1"
+          pip install uv pytest && uv pip install --system ".[dev,dask]"
 
       - name: Install lightgbm on macOS
         if: matrix.os == 'macos-latest'
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,121 @@
 
 Bug fixes and features are added through pull requests (PRs).
 
+## Prerequisites
+
+Before contributing, ensure you have:
+
+* A GitHub account
+* Basic understanding of git and GitHub
+* `uv` package manager installed (see [installation guide](https://docs.astral.sh/uv/getting-started/installation/))
+
+## Git Fork-and-Pull Workflow
+
+We use a fork-and-pull workflow for contributions. Here's how to get started:
+
+### 1. Fork and clone the repository
+
+1. Fork the repository on GitHub to your account
+2. Clone your fork locally:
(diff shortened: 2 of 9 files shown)


mlforecast/auto.py
nasaul: You need to add the docstring for `reuse_cv_splits`

nasaul (commented):
Good implementation @janrth! Just minor comments to solve before merging.

Issue #520, opened 2025-10-24
AutoSeasonalityAndDifferences crashes the kernel under certain scenarios


elephaint:
### What happened + What you expected to happen

The following code causes a kernel crash on Ubuntu (WSL) and Windows 11, both Python 3.11, on a fresh uv environment.

The kernel doesn't crash when reducing `max_diffs` inside `AutoSeasonalityAndDifferences` to `4`. I'm guessing an illegal memory issue (ptr to illegal/unexisting memory address)

### Versions / Dependencies

- Python 3.11
- Ubuntu 22.04 WSL
- Windows 11 Pro

### Reproduction script

```
import lightgbm as lgb
from xgboost import XGBRegressor
from mlforecast import MLForecast
from mlforecast.lag_transforms import ExpandingMean, RollingMean,ExponentiallyWeightedMean
from mlforecast.target_transforms import AutoSeasonalityAndDifferences, LocalStandardScaler
from utilsforecast.data import generate_series

series_count=100
params={
  "model_frequency":"MS",
  "model_horizon":12,
  "season":12,
  "eval_interval":3
}

training_data = generate_series(series_count, min_length=15, max_length=120, freq=params['model_frequency'], with_trend=True, static_as_categorical=False)

ml_models_list = [
lgb.LGBMRegressor(random_state=0, verbosity=-1),
XGBRegressor( n_estimators=500, learning_rate=0.05)
]
date_features = ['month']

ml = MLForecast(
    models=ml_models_list,
    freq=params['model_frequency'],
    lags=[1, 3],
    lag_transforms={
    1: [ExpandingMean()],
    12: [RollingMean(window_size=params['model_horizon'])],
    12: [ExponentiallyWeightedMean(alpha=0.4)],
    },
    date_features= date_features,
    target_transforms=[ LocalStandardScaler (),
              AutoSeasonalityAndDifferences(max_diffs=12,
                            max_season_length=params['season'],
                            n_seasons=3)
              ]
    )
ml.fit(training_data)
```

### Issue Severity

None

PR #547, opened 2026-01-08
Safeguard for AutoSeasonalityAndDifferences

nasaul:
Adds AutoSeasonalityAndDifferences safeguard for parameters combination that will crash the kernel due to memory issues in C++.

Solves #520 

- [x] Adds minimum data safeguards for calling AutoSeasonalityAndDifferences 
- [x] Adds new testing for the safeguard.


elephaint (approved):
Nice!

Issue #194, opened 2023-08-30
Lag transforms on different levels of aggregation


jmoralez:
### Discussed in https://github.com/Nixtla/mlforecast/discussions/192

<div type='discussions-op-text'>

<sup>Originally posted by **elisevansartefact** August 30, 2023</sup>
If I want to use lags of y for each unique_id, and lag transformations such as rolling mean per unique_id, this is easy enough to do:

`MLForecast(models = [MODELS], lags = [1, 2], lag_transforms={1: [(rolling_mean, 4)]})`

However, in order to include insights about the global trend, what if I want to include features which are transformations of the lags of every unique_id (not just one unique_id)?

For instance, what if I want to include the 4-week rolling mean of the sum of y for every unique_id? Or the mean of all the y_lag1 for each unique_id?

I suppose that this can be done via MLForecast.preprocess, however perhaps I am missing an easier way, and if not, perhaps a global_lag_transforms could be added in future as a parameter of MLForecast? </div>

PR #551, opened 2026-01-21
[FEAT] Use window functions on higher aggregation levels

janrth:
This is an addition to the lag transform functionality by allowing either global aggregation of the target value or by defining a list of categorical variables (e.g.: brand, commodity group) where the selected window functions are calculated on. One constraint here is that unique_ids should have the same ending. If unique_ids don't have different endings, then we assume using global or groupby window functions are not so meaningful. Plus, the computation time could be an issue when allowing different endings.

Usage looks like this:
RollingQuantile(p=0.5, window_size=3, groupby=["brand"])

or like this: 
fcst = MLForecast(
    models=[],
    freq='D',
    lags=[1],
    lag_transforms={1: [RollingMean(window_size=2, global_=True),
                        RollingStd(window_size=2, global_=True),
                        ExpandingMax(),
                        ExpandingMean()
                        ]},)

## Description
Allows lag transform over multiple unique_ids as discussed in:
#194 and #192

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ Documentation = "https://nixtlaverse.nixtla.io/mlforecast/"
 
 
 [dependency-groups]
-dask = ["fugue[dask]>=0.9.4", "dask[complete]<=2024.12.1", "lightgbm", "xgboost"]
+dask = ["fugue[dask]>=0.9.4", "dask[complete]", "lightgbm", "xgboost"]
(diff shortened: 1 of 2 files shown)


janrth:
@codex

janrth:
@codex

janrth:
@codex

PR #615, opened 2026-04-05
[BUG] Fix pooled ExpandingMean state init for refit=False CV and AutoML

janrth:
This fixes a bug where pooled lag transforms such as ExpandingMean(groupby=[...]) failed in MLForecast.cross_validation(..., refit=False) and AutoMLForecast.fit(..., refit=False).

The issue was in the shared predict(new_df=...) path. When refit=False, mlforecast rebuilds a TimeSeries from historical data and then continues with update-based prediction. That path was only initializing local lag transforms, but not pooled global/grouped transforms. As a result, stateful pooled transforms could reach update(...) without their internal state being materialized first.

What changed

- Added a helper in TimeSeries to initialize lag-transform state for:
  - local transforms
  - global pooled transforms
  - grouped pooled transforms
- Replaced the previous local-only initialization in MLForecast.predict(..., new_df=...) with that shared helper

This keeps the fix narrow and targets the exact path used by refit=False.

Tests
Added regressions for:

- MLForecast.cross_validation(..., refit=False) with grouped ExpandingMean
- AutoMLForecast.fit(..., refit=False) with grouped ExpandingMean

Solves #614 

Checklist:
- [ ] The tests pass.
- [ ] All linting tasks pass.


nasaul (approved):
LGTM, just removed the assertion in code and improved the testing suite.

PR #625, opened 2026-04-16
[FEAT] Add date_features_as_dummies for one-hot encoding of date features

### LINKED ISSUE #315: [date features]: dayofweek_cat - day of week as a one hot encoding feature
### Description

In my search through the documentation for the `dayofweek` parameter usage in `date_feature` inputs, I noticed that `dayofweek` is treated as an ordinal feature. However, for models such as linear regression, representing this as a one-hot encoding feature could be more efective.

Here are some suggestions I've considered:

* Introducing `dayofweek_cat` as a recognized string for creating this feature.
* Compiling a list of all acceptable strings for the `date_feature` parameter in the documentation, which seems to be missing currently.
* Including an example in the [`Custom Date Features`](https://nixtlaverse.nixtla.io/mlforecast/docs/how-to-guides/custom_date_features.html) section on how to effectively implement this. The current documentation mentions the function `is_monday`. This might lead a user to wonder whether it's possible to create a function that returns a DataFrame with multiple columns (for instance, `is_monday`, `is_tuesday`, ..., `is_sunday`). However, I validated this is not possible :(

### Use case

A user wants to:
* Use one hot encoding easily.
* Read the list of available date attributes.
* Implement custom functions that return more than one new column.
* Understand the limits of the Custom Date Features.

nasaul:
## Summary

Closes #315.

Adds a `date_features_as_dummies: bool = False` parameter to `MLForecast` and `TimeSeries`. When set to `True`, string date features with a known finite range (e.g. `dayofweek`, `month`, `quarter`) are expanded into binary indicator columns named `{feature}_{value}` instead of being kept as ordinal integers. This is particularly useful for linear models where treating cyclic calendar features as ordinal produces incorrect assumptions.

## Example

``` python
from sklearn.linear_model import Ridge

fcst = MLForecast(
    models=[Ridge()],
    freq="D",
    lags=[1],
    date_features=["dayofweek", "month", "year"],
    date_features_as_dummies=True,
)
fcst.fit(series)
```

`preprocess()` now yields columns `dayofweek_0` … `dayofweek_6`, `month_1` … `month_12`, plus `year` as an ordinal (since `year` is not in the supported set).

## Supported features for dummy encoding

| Feature | Columns | Backend |
|---|---|---|
| `dayofweek`, `day_of_week`, `weekday` | 7 | narwhals |
| `month` | 12 | narwhals |
| `quarter` | 4 | narwhals (derived from `month`) |
| `day` | 31 | narwhals |
| `hour` | 24 | narwhals |
| `minute`, `second` | 60 | narwhals |
| `dayofyear`, `day_of_year` | 366 | narwhals (`ordinal_day`) |
| `week`, `weekofyear` | 53 | pandas/polars fallback — narwhals has no equivalent |

Features outside this set (e.g. `year`) remain ordinal regardless of the flag.

## Implementation notes

- Most features are computed via **narwhals**, making the code backend-agnostic (pandas and polars).
- `week` / `weekofyear` use a backend-specific fallback (`dt.isocalendar().week` for pandas, `dt.week()` for polars) because narwhals has no equivalent datetime method.
- `_compute_date_feature` now returns `Dict[str, Any]` instead of a `(name, values)` tuple, which also unlocks support for callable date features that return a `pd.DataFrame` with multiple columns.
- For the polars path, non-dummy features keep the existing expression-based approach; dummies are computed eagerly on unique dates and joined back.

## Test plan

- `test_date_feature_dummies_pandas` — binary columns on pandas DataFrame
- `test_date_feature_dummies_polars` — same on polars DataFrame
- `test_date_feature_dummies_mixed` — dummy + ordinal features coexist correctly
- `test_date_feature_dummies_predict` — end-to-end `fit` + `predict`
- `test_date_feature_dummies_quarter` — derived feature (quarter from month) correct
- `test_date_feature_dummies_month_values` — exactly one indicator is 1 per row
- `test_date_feature_dummies_all_supported` — smoke test for every supported feature
- `test_date_feature_dummies_false_default` — default behavior unchanged


JQGoh (commented):
Should we implement this feature for DistributedMLForecast as well? I notice that it does use TimeSeries class as well.

JQGoh (approved):
LGTM

PR #627, opened 2026-04-17
[FIX] Removes pandas conversion in Distributed predictions

### LINKED ISSUE #507: Disttributed MLForecast needs a pandas data frame for predict when using exog features

The DistributedMLForecast feature is to allow the development of scalable models, we have to convert the exog features to pandas before the predict functionality, then its sure to run out of memory. Please help avoid the conversion to pandas in the predict

### Versions / Dependencies

mlforecast version `1.02`

### Reproduction script

This exact code does not work if the soze pf the data goes up [The example code on MLforecast page](https://nixtlaverse.nixtla.io/mlforecast/docs/getting-started/quick_start_distributed.html)

### Issue Severity

None

nasaul:
## Summary

`DistributedMLForecast.predict()` previously required `X_df` to be a pandas `DataFrame`, which forced callers to materialise all future exogenous features on a single node — defeating the purpose of distributed forecasting.

This PR lifts that restriction so `X_df` can be any distributed DataFrame (Dask, Spark, Ray, or pandas). The implementation avoids sending the full `X_df` to every worker by joining and packing it per partition before prediction.

Closes #507 

### Changes

**Distributed `X_df` support**
- Removed the `isinstance(X_df, pd.DataFrame)` guard; `X_df` now accepts `fugue.AnyDataFrame`.
- Added `_build_x_df_per_partition()`: builds a `uid → partition` map on the driver, distributes the join back to workers, and serialises each partition's `X_df` slice as a base64 blob so it survives Dask/Spark type inference.
- Added `_pack_x_df()` and `_attach_x_df()` helpers. `_attach_x_df` uses `fa.transform` instead of `fa.join` to avoid Fugue's Ray engine routing joins through DuckDB (which materialises results locally and breaks distributed execution).

**Spark partitioning fix**
- Added an explicit `partition=dict(by=id_col)` for Spark `DataFrames` in `_preprocess_partitions`. Without it, rows for the same series can land in different Spark partitions, producing duplicate or incomplete predictions.

**Schema and field access refactor**
- Introduced `_PARTITION_FIELDS` and named-field access (`dict(zip(...))`) throughout, replacing brittle positional tuple unpacking. Added `first_uid` (str) and `all_uids` (binary) columns to the partition schema to support the distributed `X_df` join without a driver-side collect.

**`date_features_as_dummies` in distributed mode**
- Propagated the `date_features_as_dummies` parameter (added in #625) to `DistributedMLForecast`, so one-hot date features are available in distributed pipelines as well.

### Testing
- Added Ray, Dask, and Spark test cases covering `predict` with a distributed `X_df`.
- Extended unit tests for `date_features_as_dummies` in distributed mode.


mlforecast/distributed/forecast.py
```diff
@@ -451,25 +465,79 @@ def fit(
...
+        yield part
+
+    @staticmethod
+    def _pack_x_df(part: pd.DataFrame, partition_key_col: str) -> Iterable[pd.DataFrame]:
+        """Serialize one partition's X_df rows as a base64 string, keyed by first_uid.
+
+        Base64 encoding avoids binary-column type issues when the packed blob flows
+        through Dask / Spark joins before reaching ``_predict``.
+        """
+        partition_key = part[partition_key_col].iloc[0]
```
gee-senbong, line 491: Do we need to worry when `part[partition_key_col]` is empty?

gee-senbong (approved):

mlforecast/distributed/forecast.py
nasaul: No, part cannot be empty here. `_pack_x_df` is invoked via `fa.transform(..., partition={"by": "__partition_key__"})`, which is a group-by partition — Fugue only calls the function for groups that exist in the data. Empty partitions only arise with count-based repartitioning, which isn't used here.

PR #632, opened 2026-04-22
[FEAT] add support for `drop_auxiliary_columns` field

### LINKED ISSUE #631: Support using groupby/partition_by columns for lag aggregations without including them as model features
## Summary

I would like to request support for using columns in `groupby` / `partition_by` lag aggregations without forcing those same columns to be part of the model input matrix.

This would help when a column is useful to define an aggregation bucket, but not desirable as a training feature.

## Motivation

A common case is wanting to compute an aggregated lag feature such as:

- `RollingMean(..., groupby=["brand"])`
- `ExpandingMean(..., partition_by=["promo"])`

but not wanting to train on:

- `brand`, because it may be very high-cardinality
- `promo`, because it may be less informative than another feature such as `discount` or `price_ratio`

In other words, these columns are useful as transformation context, but not necessarily as model features.

## Current limitation

Today, the workaround is to call `preprocess`, manually drop the columns, and then call `fit_models`.

That works for a manual train flow, but it breaks the equivalent workflow in `cross_validation`, since `cross_validation()` internally handles preprocessing/training and there is no way to say:

- use `brand` / `promo` to compute lag aggregations
- but exclude them from the feature matrix passed to the model

This makes it difficult to evaluate these transformations consistently during backtesting.

## Example

Suppose I want:

- a rolling mean aggregated by `brand`
- a partitioned lag transform by `promo`
- to train only on the engineered lag features plus other exogenous variables such as `discount` / `price_ratio`

For example:

```python
lag_transforms = {
    1: [
        RollingMean(7, groupby=["brand"]),
        ExpandingMean(partition_by=["promo"]),
    ]
}
```

## Desired behavior

- `brand` is used to build the grouped lag feature
- `promo` is used to build the partitioned lag feature
- neither `brand` nor `promo` is included in the final model feature matrix

## Why this matters

This is especially important in `cross_validation()`, where there is currently no clean way to reproduce the "preprocess, drop columns, then train" flow.

It would also be helpful in regular `fit()` / `predict()` flows, because these columns may still be required to compute the transforms, even if they should not be exposed to the model.

## Static feature consideration

There is also an interaction with `static_features`.

For example, `groupby=["brand"]` currently requires `brand` to be available as a static feature, which makes sense for computing the aggregation. However, I may still want `brand` to be dropped before the model `fit()` step.

So it would be useful to distinguish between:

- columns required internally to compute transforms
- columns that should actually be included in the model feature matrix

## Possible API directions

I’m not attached to a specific API, but a few possibilities would be:

- `drop_features=[...]`
- `exclude_features=[...]`
- `transform_only_features=[...]`
- a separate distinction between:
  - features needed for transform computation
  - features used for model training

The important part is that the same behavior should work consistently in:

- `fit`
- `predict`
- `preprocess`
- `cross_validation`

## Expected semantics

Ideally, columns marked as transform-only would:

- still be available to compute `groupby` / `partition_by` lag features
- still be accepted in future `X_df` when needed for recursive prediction
- not be included in the feature matrix passed to the model
- not need the user to manually split the pipeline into `preprocess` + `fit_models`
- behave consistently in `cross_validation`

## Additional note

This would be especially useful for high-cardinality static columns and low-signal partition columns that are necessary for feature engineering but undesirable as direct model inputs.

simonez-tuidi:
## Description

Adds a `drop_auxiliary_columns` parameter to `TimeSeries` and `MLForecast` to control which columns are excluded from the model feature matrix after feature engineering.

> **⚠️ Behavior change:** The default `True` will silently drop any column used as a `groupby` in a lag transform from the feature matrix. If you rely on those columns being passed to your model, set `drop_auxiliary_columns=False` to restore the previous behavior.

#### `mlforecast/core.py`
- Added `drop_auxiliary_columns: Union[bool, Sequence[str]] = True` to `TimeSeries.__init__`.
- Updated `_fit()` so that, after `features_order_` is built, columns are excluded from it according to the parameter value.
- Emits a `UserWarning` if a requested column name is not present in the generated feature set (only when an explicit list is passed).

#### `mlforecast/forecast.py`
- Added `drop_auxiliary_columns` to `MLForecast.__init__` and passed it through to `TimeSeries`.
- Propagated the parameter when constructing `new_ts` inside `predict(new_df=...)` and `_compute_recursive_fitted_values_on_demand`.

### Behavior

`drop_auxiliary_columns` accepts three forms:

| Value | Effect |
|---|---|
| `True` (default) | Drops all columns referenced in any `groupby` of a lag transform |
| `False` | Keeps all columns (previous behavior) |
| `["col1", "col2"]` | Drops exactly the named columns; warns if any are not found |

`features_order_` is the single source of truth for what gets passed to models. Filtering it means dropped columns are excluded in `fit`, `predict`, `preprocess(return_X_y=True)`, and `cross_validation`, while remaining available in `static_features_` so groupby transform computation still works.

### Tests

Added 6 tests in `tests/test_pipeline.py` covering:
- default `True` auto-dropping groupby columns while keeping computed lag features
- `False` keeping groupby columns in the feature matrix
- explicit list dropping exactly the named columns
- dropped columns being absent from the prediction feature matrix (via `FeatureRecorder`)
- consistent behavior in `cross_validation`
- `UserWarning` emission for unknown column names

Closes #631

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.


simonez-tuidi:
@nasaul might need a hand with the docs, I couldn't run the `make all_docs` command locally as I ran into this error:

```
mkdir -p nbs/_extensions
cp -r docs-scripts/mintlify/ nbs/_extensions/mintlify
python docs-scripts/update-quarto.py
quarto render nbs/docs --output-dir ../docs/mintlify/
ERROR: Unsupported project type mintlify

Stack trace:
    at projectType (file:///opt/quarto/1.4.515/bin/quarto.js:37823:15)
    at projectContext (file:///opt/quarto/1.4.515/bin/quarto.js:72232:30)
    at eventLoopTick (ext:core/01_core.js:183:11)
    at async render (file:///opt/quarto/1.4.515/bin/quarto.js:81266:19)
    at async Command.fn (file:///opt/quarto/1.4.515/bin/quarto.js:81414:32)
    at async Command.execute (file:///opt/quarto/1.4.515/bin/quarto.js:8095:13)
    at async quarto (file:///opt/quarto/1.4.515/bin/quarto.js:113977:5)
    at async file:///opt/quarto/1.4.515/bin/quarto.js:113995:9
make: *** [Makefile:13: examples_docs] Error 1
```

nasaul:
Have you already installed [mintlify](https://github.com/Nixtla/mlforecast/blob/main/CONTRIBUTING.md#install-mintlify)?

simonez-tuidi:
Here's a summary of all changes made:

- mlforecast/core.py
Added Sequence to typing imports
Replaced drop_features: Optional[List[str]] = None with drop_auxiliary_columns: Union[bool, Sequence[str]] = True
Replaced self.drop_features with self.drop_auxiliary_columns
Replaced the explicit-list filtering block with three-branch logic: True auto-collects all groupby columns across transforms, False is a no-op, and a sequence drops the named columns explicitly (with the unknown-column warning preserved for that case only)

- mlforecast/forecast.py
Added Sequence to typing imports
Renamed the parameter and updated the docstring in __init__
Propagated drop_auxiliary_columns in all three TimeSeries(...) construction sites: __init__, predict(new_df=...), and cross_validation

- tests/test_pipeline.py
Replaced the five old test_drop_features_* tests with six updated ones covering: default auto-drop, explicit False opt-out, explicit sequence, predict exclusion, cross-validation, and unknown-column warning

nasaul (approved):
Thanks for your contribution @simonez-tuidi :)

PR #641, opened 2026-05-08
[FIX] Pooled global/groupby lag transforms to use RANGE semantics

### LINKED ISSUE #640: [CORE] `groupby`/`global_` transforms use sum-then-roll semantics instead of SQL RANGE window semantics
## Summary

The existing `groupby` and `global_` transform paths (introduced in #551) aggregate multi-series observations by timestamp using `sum` before applying the rolling window. This produces results that diverge from SQL `RANGE BETWEEN ... PRECEDING` semantics and scale with the number of series in a group, which limits their usefulness in practice.

## Current behavior

When multiple series share a group (via `groupby=["brand"]` or `global_=True`), the implementation in `core.py` line 917-920 collapses all series at each timestamp into a single value using `group_by_agg(..., {target_col: "sum"})`. The rolling transform then operates on these summed-per-timestamp values.

### Example: `RollingMean(window=2, lag=1, global_=True)`

Given two series:

| ds | a (y) | b (y) |
|----|-------|-------|
| 1  | 1     | 10    |
| 2  | 2     | 20    |
| 3  | 3     | 30    |
| 4  | 4     | 40    |

**Step 1 — sum by timestamp:**

| ds | y_sum |
|----|-------|
| 1  | 11    |
| 2  | 22    |
| 3  | 33    |
| 4  | 44    |

**Step 2 — RollingMean(window=2, lag=1) on the summed series:**

| ds | result          |
|----|-----------------|
| 1  | NaN             |
| 2  | NaN             |
| 3  | mean(11, 22) = **16.5** |
| 4  | mean(22, 33) = **27.5** |

This matches [the existing test expectation](https://github.com/Nixtla/mlforecast/blob/main/tests/test_core.py#L646).

## Expected behavior (SQL RANGE semantics)

The equivalent SQL window function would be:

```sql
AVG(y) OVER (RANGE BETWEEN 2 PRECEDING AND 1 PRECEDING)
```

This takes the mean of **all individual observations** in the window, regardless of how many series share a timestamp:

| ds | result                           |
|----|----------------------------------|
| 1  | NaN                              |
| 2  | NaN                              |
| 3  | mean(1, 2, 10, 20) = **8.25**   |
| 4  | mean(2, 3, 20, 30) = **13.75**  |

## Why this matters

1. **Results scale with group size.** The sum-then-roll approach means that adding more series to a group changes the rolling mean proportionally. A `RollingMean` should produce values in the same range as the target regardless of how many series contribute — sum-then-roll breaks this invariant.

2. **Results change with changes to the number of series belonging to a group.** As often happens in typical retail scenarios, inventories can change and so do the number of products belonging to a certain category. Since the rolling mean is proportional to the number of series in the group, this would lead to changes in the values of the average which would likely produce sudden drops/jumps and a "jagged" profile, which is inconsistent with the expected behavior of a mean aggregation which should lead to smoother results.

3. **Inconsistency with `partition_by`.** The new `partition_by` code (PR #636) correctly implements RANGE semantics — individual observations are preserved and the rolling window sees all raw values. Having `groupby` use sum-then-roll while `partition_by` uses RANGE makes the two features produce inconsistent results for equivalent configurations.

4. **The test with one series per group masks the issue.** The `test_group_lag_transform` test (line 662) assigns brand "x" to series "a" and brand "y" to series "b", so each group contains exactly one series. The sum aggregation is a no-op and results coincide with per-series values, hiding the semantic difference.

## SQL equivalence model

The natural mental model for these transforms maps cleanly to SQL window functions:

| mlforecast configuration | SQL equivalent |
|---|---|
| `RollingMean(w, lag=l)` (default) | `AVG(y) OVER (PARTITION BY unique_id RANGE BETWEEN ...)` |
| `RollingMean(w, lag=l, groupby=["brand"])` | `AVG(y) OVER (PARTITION BY brand RANGE BETWEEN ...)` |
| `RollingMean(w, lag=l, partition_by=["promo"])` | `AVG(y) OVER (PARTITION BY unique_id, promo RANGE BETWEEN ...)` |
| `RollingMean(w, lag=l, partition_by=["promo"], groupby=["brand"])` | `AVG(y) OVER (PARTITION BY brand, promo RANGE BETWEEN ...)` |

In all cases the transform should find all observations that belong to the partition and fall within the RANGE window, then apply the statistic.

### Option A: Change default to RANGE semantics (breaking change)

Switch `groupby`/`global_` to keep individual rows (like `partition_by` already does) so rolling transforms see raw observations. The per-series default and `partition_by` already follow RANGE semantics, making this the consistent choice.

This changes numeric outputs for users of `groupby`/`global_` transforms with multiple series per group.

### Option B: Add an `agg` parameter (non-breaking)

Introduce an `agg` parameter on rolling transforms:

```python
RollingMean(
    window=2,
    groupby=["brand"],
    agg=None,    # None = RANGE semantics (mean of all individual observations)
                 # "sum" = current behavior (sum by timestamp, then roll)
                 # "mean" = mean by timestamp, then roll
)
```

Default to `agg=None` (RANGE) for new code, but allows `agg="sum"` to reproduce the current behavior for backward compatibility.

**Note on commutativity:** For `agg="mean"`, aggregating per-timestamp first and then applying a rolling mean across timestamps does NOT produce the same result as RANGE semantics when timestamps have different numbers of observations. Consider the partition example:

| ts | a (y) | b (y) |
|----|-------|-------|
| 1  | 10    | --    |
| 2  | 15    | --    |
| 3  | 12    | 20    |
| 4  | --    | 23    |

- RANGE mean over all 4 observations: (10 + 15 + 12 + 20) / 4 = **14.25**
- Aggregate-then-roll: mean of per-ts means = mean(10, 15, (12+20)/2) = mean(10, 15, 16) = **13.67**

These differ because timestamps 1-2 have one observation while timestamp 3 has two. The `agg` parameter makes this explicit rather than leaving it as a hidden assumption.

### Recommendation

Option A is simpler and produces correct results. The current sum-then-roll behavior has limited practical value and the existing tests with one series per group suggest it wasn't thoroughly validated for multi-series groups. Option B adds complexity to preserve a behavior that may not have real users.

## Affected code

- `mlforecast/core.py` lines 917-920: `group_by_agg(..., {target_col: "sum"})` in `fit_transform`
- `mlforecast/core.py` lines 2118-2121: same aggregation in `update`
- `mlforecast/core.py` lines 1063-1089: group transform application and join-back (uses `group_cols + [time_col]` join key, would need `[id_col, time_col]` for RANGE)
- `tests/test_core.py` line 646: `test_global_lag_transform` expected values
- `tests/test_core.py` line 691: `test_group_lag_transform` expected values (unaffected when one series per group, but should add multi-series-per-group test)

### Versions / Dependencies

1.0.31

### Reproduction script

See above

### Issue Severity

None

simonez-tuidi:
# PR Description

## Summary

This PR reworks `global_` and `groupby` lag transforms so pooled features are computed over the underlying observations in each time range, matching SQL-style `RANGE BETWEEN ... PRECEDING` semantics.

This implements **Option A** from the issue #640 : change the default `global_` / `groupby` behavior to RANGE semantics instead of preserving the current sum-then-roll behavior.

Previously, global and grouped transforms were backed by separate ad hoc state paths that aggregated each timestamp using `sum` before applying the lag transform. That made transforms such as:

```python
RollingMean(window_size=2, global_=True)
RollingMean(window_size=2, groupby=["brand"])
```

behave like transforms over per-timestamp sums rather than transforms over all rows in the relevant time window. This branch introduces a shared pooled state representation and computes pooled transforms directly from bucketed observation arrays.

## Problem Addressed

When multiple series share a pooled bucket, the old implementation first collapsed the data by timestamp:

| ds | a | b | summed y |
| --- | ---: | ---: | ---: |
| 1 | 1 | 10 | 11 |
| 2 | 2 | 20 | 22 |
| 3 | 3 | 30 | 33 |
| 4 | 4 | 40 | 44 |

Then `RollingMean(window_size=2, lag=1, global_=True)` operated on `[11, 22, 33, 44]`, producing values such as `mean(11, 22) = 16.5`.

That has two practical problems:

- `RollingMean` scales with the number of series in the group, because it is effectively averaging sums.
- Results can jump when the membership of a group changes, even if the target distribution is stable.

With this PR, the same transform operates over the individual observations in the RANGE window. For example, at `ds=3` the window contains `[1, 10, 2, 20]`, so the mean is `8.25`.

One detail worth making explicit: `min_samples` is still evaluated over observations. With multiple series in a bucket, a window containing one timestamp can satisfy `min_samples=2` if that timestamp has two observed rows.

### Added `PooledState`

Added `mlforecast/pooled.py` with a `PooledState` object that owns the state needed by pooled transforms:

- flat observation arrays: bucket id, timestamp, time index, and target value
- a `GroupedArray` for existing transform state initialization
- bucket metadata used to join computed features back to the original dataframe
- group key mappings for `groupby` transforms
- update logic for predictions, new observations, new groups, and new series

This replaces the previous separate `_global_ga` / `_global_times` and `_group_states` code paths with a single state model.

### Compute Pooled Features Directly

Lag transforms already expose `_compute_bucket_feature(...)` (added in earlier commits on this branch) for rolling, seasonal, expanding, EWM, `Offset`, and `Combine` transforms. This PR routes all pooled computation through those methods via a new `compute_pooled_features()` function, and removes the previous silent fallback to positional `GroupedArray` behavior.

The old fallback was problematic: when a transform did not implement `_compute_bucket_feature`, the code silently fell back to GA positional semantics, which produced incorrect results under RANGE window bounds. Unsupported pooled transforms now raise a clear `NotImplementedError` instead.

### SQL-Like Range Semantics

Pooled transforms now use a per-bucket `time_index` derived from the validated regular time grid. For global and groupby transforms, this gives interval-style window bounds while preserving the existing codebase assumption that non-partitioned series do not contain gaps.

This means a grouped feature behaves like:

```sql
AVG(y) OVER (
  PARTITION BY brand
  RANGE BETWEEN 2 PERIODS PRECEDING AND 1 PERIOD PRECEDING
)
```

rather than first aggregating `y` by `(brand, timestamp)`.

The intended equivalence model is:

| mlforecast configuration | SQL mental model |
| --- | --- |
| `RollingMean(w, lag=l)` | `AVG(y) OVER (PARTITION BY unique_id RANGE BETWEEN ...)` |
| `RollingMean(w, lag=l, global_=True)` | `AVG(y) OVER (RANGE BETWEEN ...)` |
| `RollingMean(w, lag=l, groupby=["brand"])` | `AVG(y) OVER (PARTITION BY brand RANGE BETWEEN ...)` |

### Update Path Fixes

The pooled state update path now keeps all related arrays and metadata in sync:

- appends observations to `bucket_df`
- recomputes/extends time indexes after `update()`
- handles new series and new groups
- updates series-to-bucket mappings after static features change
- preserves prediction-time feature computation after updates

This also fixes a pre-existing bug in `TimeSeries.update()` where new series received wrong static features: `ufp.take_rows(df, ...)` was indexing into the full DataFrame instead of the new-series subset, causing incorrect bucket assignments for series introduced via `update()`.

### Categorical Group Key Support

Grouped buckets are represented internally by numeric `_bucket_id`s, but public group keys such as `brand` or `subcategory` may be categorical. The new helpers reconcile pandas and Polars categoricals before joins and concatenations, including when updates introduce a new group value.

## Tests

Added `tests/test_pooled.py` covering:

- global and grouped update state preservation
- sequential updates
- staggered series starts
- new series in new groups
- categorical group keys with new group values
- prediction-time feature computation after updates
- unsupported pooled transform errors

Updated existing core tests to assert the new RANGE-style semantics for global and grouped rolling/expanding transforms.

The previous `test_group_lag_transform` used one series per group, which made sum-by-timestamp and RANGE semantics indistinguishable. The updated tests include multiple series in the same group so this behavior is covered directly.

## Verification

Ran:

```bash
python -m pytest tests/test_pooled.py -x -q
python -m pytest tests/test_core.py -x -q
ruff check mlforecast/core.py mlforecast/pooled.py mlforecast/lag_transforms.py tests/test_pooled.py tests/test_core.py
```

The full branch suite was also verified with:

```bash
python -m pytest tests/test_core.py tests/test_lag_transforms.py tests/test_forecast.py tests/test_pooled.py -x -q
```

Result:

- 279 passed
- 2 skipped
- lint clean
- `mlforecast/pooled.py` at 98% coverage

## Compatibility / Breaking Change

The public transform API is unchanged. Existing `global_` and `groupby` arguments continue to be used.

This is nevertheless a necessary breaking change for users who already rely on `global_` or `groupby` transforms with more than one series in a pooled bucket. The numeric output changes from "sum by timestamp, then apply the transform" to "apply the transform over all observations in the RANGE window".

For example, `RollingMean(window_size=2, lag=1, global_=True)` over two aligned series changes from:

```text
ds=3: mean(sum(ds=1), sum(ds=2)) = mean(11, 22) = 16.5
```

to:

```text
ds=3: mean(all observations at ds=1 and ds=2) = mean(1, 10, 2, 20) = 8.25
```

This change is intentional because the previous behavior made means scale with the number of series in the group and diverged from the SQL `RANGE` mental model used by the rest of the pooled/partitioned transform design. Preserving the old behavior would require adding a separate aggregation mode (for example, "sum by timestamp before transforming"), which would keep the incorrect default and add API complexity. This PR chooses correctness and consistency instead.

Users who depended on the old sum-then-transform behavior will need to reproduce that aggregation explicitly before fitting or use a future explicit aggregation option if one is added.

The internal fitted `TimeSeries` state shape changed, so previously pickled fitted `TimeSeries` objects that depend on the old private pooled state are not expected to be compatible.

## Linking issue

Closes #640


janrth:
Will this help finding a smarter solution for the partition_by problem?

As I just mentioned on the issue itself, I still like option B (from the issue) as I see reasons why we would first do the sum over ds before rolling window functions. 

For example, check out this entry in discussion:
https://github.com/Nixtla/mlforecast/discussions/644

simonez-tuidi:
@janrth 

Thanks Jan, yes I believe this approach would help bridge the gap between `groupby` in the current implementation and the proposed `partition_by` logic - by introducing the `PooledState` object that is tasked with handling both - actually all three, namely `global` or `groupby` plus `partition_by`

I'll try and wrap up soon with the new PR which branches off of this branch and includes the `partition_by` implementation using `PooledState`

Regarding the issue just raised, I believe this would work nicely with either implementation, and in my use cases I would see it more beneficial to go with Option A - using `Combine` to take the ratio of two RollingMean, one across a single series and one grouped or global

janrth:
@codex

mlforecast/pooled.py
```diff
@@ -0,0 +1,547 @@
...
+    groups_copy = groups
+    for col in cols:
+        s1, s2 = ufp.match_if_categorical(data_slice[col], groups_copy[col])
+        data_slice = data_slice.with_columns(s1)
+        groups_copy = groups_copy.with_columns(s2)
+    joined = data_slice.join(groups_copy, on=cols, how="left")
+    return joined["_bucket_id"].to_numpy()
+
+
+def _compute_time_index(bid_arr, ts_arr):
```
janrth, line 104: Since pooled `global_` / `groupby` transforms assume a continuous calendar, can we warn when `validate_data=False` and the model contains pooled transforms?

The public validation path catches gaps, but disabling validation allows pooled time indexes to compress observed timestamps. For example, ds=[1, 3] becomes adjacent internal periods, so a lag-1 rolling feature at ds=3 can use ds=1 even though freq=1 would imply ds=2.

Maybe add a helper like `_warn_if_pooled_tfms_without_validation()` and call it when `validate_data=False` in preprocess/fit/cross_validation.

simonez-tuidi:
@nasaul @janrth 

One thing worth noting here which wasn't included in the PR description: the new logic treats the `min_samples` parameter in quite a literal way, which I thought was the better option but could be up for interpretation. In the current code, the samples represent the number of individual observations, whereas in the previous implementation of `groupby` we were counting the number of individual timestamps that form part of the window.

So now if we perform a `RollingMean(window_size=1, min_samples=2, groupby=["brand"])` the result will be not null if there are at least two series.

Now this behavior diverges from the **single series** implementation, where min_samples is always capped at window_size if the user passes a higher value.

I'm going to add better docs to this branch and explain this behavior more explicitly, pending discussion.

In my opinion we could add in a separate feature branch an optional parameter `min_intervals` which counts the number of observed timestamps with non-null values across the aggregation, what do you think?

mlforecast/pooled.py
```diff
@@ -0,0 +1,547 @@
...
+    for bid in np.unique(bid_arr):
+        mask = bid_arr == bid
+        ts_b = ts_arr[mask]
+        unique_ts = np.unique(ts_b)
+        idx_arr[mask] = np.searchsorted(unique_ts, ts_b)
+        next_by_bucket[int(bid)] = len(unique_ts)
+    return idx_arr, next_by_bucket
+
+
+@dataclass
```
janrth, line 124: I think my main concern is currently around performance.

Using global_=True on a data set with 20k unique_ids showed up to 10 times slower performance than global_=False (I tested with different lengths for ds, and the longer the ds, the bigger the difference got).

But the main slowdown was not the rolling mean calculation itself, but doing global pooled work at row level.

I have tested a few iterations and found that a good improvement can be made by keeping the current pooled/raw-row semantics but caching compact per-timestamp aggregates in `PooledState`:

- `sums_by_time`
- `counts_by_time`
- `unique_times`
- whether the global panel is balanced

Then global rolling mean is computed over the compact time axis instead of over all `(unique_id, ds)` rows. For balanced panels, the result is broadcast back with `np.tile` instead of a row-level merge/search.

This avoids repeated row-level `bincount`, `np.unique`, and timestamp mapping for the common balanced-panel case. Recursive prediction also uses the cached per-time aggregates to compute the next global mean directly, instead of building full query arrays.

In my benchmark this reduced the global rolling mean runtime substantially while preserving the new pooled semantics.

If you want I can push my changes into your branch, but I am sure your agent can come to similar results, when shown the general idea.

janrth:
> @nasaul @janrth
> 
> One thing worth noting here which wasn't included in the PR description: the new logic treats the `min_samples` parameter in quite a literal way, which I thought was the better option but could be up for interpretation. In the current code, the samples represent the number of individual observations, whereas in the previous implementation of `groupby` we were counting the number of individual timestamps that form part of the window.
> 
> So now if we perform a `RollingMean(window_size=1, min_samples=2, groupby=["brand"])` the result will be not null if there are at least two series.
> 
> Now this behavior diverges from the **single series** implementation, where min_samples is always capped at window_size if the user passes a higher value.
> 
> I'm going to add better docs to this branch and explain this behavior more explicitly, pending discussion.
> 
> In my opinion we could add in a separate feature branch an optional parameter `min_intervals` which counts the number of observed timestamps with non-null values across the aggregation, what do you think?

I realised this and had to think about this for a minute. I think totally fine. If people do a detailed deep dive into preprocess and inspect data they might realise the difference, but it is logically consistent and in line with the sql notation.

Making a few more doc strings will do the job for me!

mlforecast/pooled.py
simonez-tuidi: You raise a good point, but I was under the impression that gaps are always raised and produce an error when instantiating the TimeSeries class, I might need to do some more digging and look into it, let me get back to you on this

simonez-tuidi:
Thanks for taking the time to review and for the feedback, really appreciate it! @janrth 

> 
> I realised this and had to think about this for a minute. I think totally fine. If people do a detailed deep dive into preprocess and inspect data they might realise the difference, but it is logically consistent and in line with the sql notation.
> 
> Making a few more doc strings will do the job for me!

mlforecast/pooled.py
janrth: I think the default setting is validate=True or so. But if people are choosing =False on purpose we should warn them that this might lead to issues in global / groupby.

simonez-tuidi:
@janrth 

I've added four commits addressing the comments you raised:

1.  ac8e7d3 - Documents that min_samples in pooled mode counts total observations across all series (no capping), unlike coreforecast's per-series behavior.

2. 0d6ede4 - Emits a UserWarning when validate_data=False and pooled transforms exist. Suppressed in cross_validation internal calls where validation is done upfront.

3. 54874ba - Adds _TimestampAggregates cache on PooledState with per-bucket sums/counts/n_rows. RollingMean uses the compact T-length arrays for both _transform() and compute_pooled_features(). Other transforms marked with TODOs.

4. 95fa716 - Adds _compute_latest_from_aggs so RollingMean during recursive prediction computes directly from cached aggregates, skipping build_query_arrays entirely. 12-17x speedup on the prediction path.

Here's the specifics of the benchmarks:

Setup: 10,000 series, 100 timestamps each, `RollingMean(28)`, 20 recursive update steps, 3 repeats. Timing isolates `ts._update_features(); ts._update_y(...)`, so it measures the changed prediction path rather than model overhead.

| Scenario | Impl | Recursive 20 steps | Per step | Speedup |
|---|---:|---:|---:|---:|
| `global_=True` | `3267a2f` | 1.705s | 85.25ms | 1.0x |
| `global_=True` | current | 0.138s | 6.92ms | 12.3x |
| `groupby=["brand"]`, 100 groups x 100 series | `3267a2f` | 5.672s | 283.59ms | 1.0x |
| `groupby=["brand"]`, 100 groups x 100 series | current | 0.330s | 16.51ms | 17.2x |

`fit_transform` time was basically unchanged, as expected: global around 0.51-0.53s, groupby around 1.07-1.12s.

I also ran a short checksum comparison for the first 3 recursive steps on both implementations; sums and NaN counts matched for both global and groupby cases.

mlforecast/core.py
```diff
@@ -704,54 +680,28 @@ def _transform(
...
+            if self._sort_idxs is not None:
+                df_sorted = ufp.take_rows(df, self._sort_idxs)
+            else:
+                df_sorted = df
         if global_tfms:
-            assert self._global_ga is not None
-            assert self._global_times is not None
-            global_vals = self._global_ga.apply_transforms(
-                transforms=global_tfms, updates_only=False
+            assert self._pooled_global is not None
```
janrth, line 707: I think the fit and predict got pretty fast, the preprocess itself is still a bit slow. 

Here is a benchmark with the way i will describe below.

current PR head:
global_preprocess: 0.333s
global_fit:        0.288s
global_predict:    0.425s

with a timestamp-level preprocess fast path:
global_preprocess: 0.217s
global_fit:        0.266s
global_predict:    unchanged / noise

The idea is to reuse the cached per-timestamp aggregates already held in PooledState._ts_aggs. For global_=True RollingMean, we do not need to scan the pooled row arrays for every row/timestamp. We can compute the rolling mean once per timestamp from cumulative sums/counts, then broadcast/map those timestamp-level values back to the original rows.

def _rolling_mean_from_aggs(sums, counts, lag, window_size, min_samples):
    csum = np.r_[0.0, np.cumsum(sums)]
    ccount = np.r_[0.0, np.cumsum(counts)]

    ords = np.arange(sums.size)
    upper = ords - lag
    lower = upper - window_size + 1

    valid = upper >= 0
    upper_clip = np.clip(upper, 0, sums.size - 1)
    lower_clip = np.clip(lower, 0, sums.size)

    window_sums = csum[upper_clip + 1] - csum[lower_clip]
    window_counts = ccount[upper_clip + 1] - ccount[lower_clip]

    out = np.full(sums.size, np.nan)
    ok = valid & (window_counts >= min_samples) & (window_counts > 0)
    out[ok] = window_sums[ok] / window_counts[ok]
    return out


Then in the global training/preprocess path:


agg = state._ts_aggs[0]

vals_by_time = _rolling_mean_from_aggs(
    sums=agg.sums,
    counts=agg.counts,
    lag=tfm._core_tfm.lag,
    window_size=tfm.window_size,
    min_samples=tfm.min_samples or tfm.window_size,
)

unique_times = np.unique(state.time)
ords = np.searchsorted(unique_times, df_sorted[time_col].to_numpy())
features[name] = vals_by_time[ords]

janrth: Unless I did not see some flaw in that logic, I think it can increase speed. I feel especially with huge data sets (+500mill. rows) speed is important (in combination with being correct, of course :) )

simonez-tuidi: Should be improved now, let me know if we can resolve this point, and thanks for raising this!

janrth (commented):
Great work!

I was wondering if we could use the same timestamp-level aggregate fast paths for min, max and std. Currently they are using the generic pooled masking path, which is slower. 

I know it is annoying and honestly I did not have that in mind earlier. But maybe we can have one more check for min, max and std. 

I think for quantile rolling we would need a more complex data structure as it needs the full distribution in the rolling window. So let's leave this out for now, but would be great if you could have a final look at min, max, std and see if they can use the fast timestamp path.

janrth:
@codex

mlforecast/core.py
janrth: Much better, I think!

mlforecast/lag_transforms.py
```diff
@@ -373,14 +688,54 @@ def __init__(
...
+                mask = ord_b == o
+                vals = y_b[mask]
+                vals = vals[~np.isnan(vals)]
+                if len(vals) > 0:
+                    mean_per_ord[k] = np.mean(vals)
+            feat_u = np.full(len(unique_ord), np.nan)
+            ewm = np.nan
+            for k in range(len(unique_ord)):
+                if unique_ord[k] > unique_ord[0] + lag - 1:
+                    feat_u[k] = ewm
```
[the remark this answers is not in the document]
janrth: @simonez-tuidi It seems for lag-2 there is a systematic issue. I think codex is right here.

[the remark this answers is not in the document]
simonez-tuidi: I'll look into this asap, weird that I used both Claude and Codex locally to cross-review and neither picked it up haha

simonez-tuidi:
> Great work!
> 
> I was wondering if we could use the same timestamp-level aggregate fast paths for min, max and std. Currently they are using the generic pooled masking path, which is slower.
> 
> I know it is annoying and honestly I did not have that in mind earlier. But maybe we can have one more check for min, max and std.
> 
> I think for quantile rolling we would need a more complex data structure as it needs the full distribution in the rolling window. So let's leave this out for now, but would be great if you could have a final look at min, max, std and see if they can use the fast timestamp path.

Thanks @janrth ! I already started implementing those and yes, for quantile we need to fall back on the "slow" implementation but for everything else it should still be solid

simonez-tuidi:
@janrth @nasaul 

I've added another sanity check by comparing results of the transforms with SQL equivalent semantics, on top of the hard-coded values in the other tests

Now all transforms that support groupby/global mode have been reimplemented and where possible are optimized for efficiency by avoiding joins and broadcasting values instead.

Let me know what you think, I'd love to see this merged soon and start implementing the `partition_by` feature on top of it

It would also be great to have it be part of the milestones in the next release, what do you think?

janrth:
> @janrth @nasaul
> 
> I've added another sanity check by comparing results of the transforms with SQL equivalent semantics, on top of the hard-coded values in the other tests
> 
> Now all transforms that support groupby/global mode have been reimplemented and where possible are optimized for efficiency by avoiding joins and broadcasting values instead.
> 
> Let me know what you think, I'd love to see this merged soon and start implementing the `partition_by` feature on top of it
> 
> It would also be great to have it be part of the milestones in the next release, what do you think?

Give ma a few days, but will look into your latest changes once I have a bit of time. But I have the feeling you have gotten pretty far and we are hopefully close to merging :)

nasaul:
Hey @simonez-tuidi this is a great job! I was ooo but great to see your work. Shout-out to @janrth for diving in with you. In order to merge I would argue that we need to work on:

- [x] Updating the docs
- [ ] Add `RollingSum` and `ExpandingSum` (Another PR)
- [x] Translate `pandas/polars` logic into `narwhals` logic. This creates an easier time for reading and maintaining code without the splits.

If you're okay I can finish this tasks for merging.

simonez-tuidi:
@nasaul thanks for that, absolutely go ahead, totally agree with switching to Narwhals to make it more maintainable

On adding `RollingSum` and `ExpandingSum` perhaps that would be best suited for merging in its own separate PR?

I can see other transforms being beneficial - like Counts - and we would also need to find the right design (e.g. do we sum all observations or is it intended to sum per-timestamp and then average?)

Another issue that was raised in #644  would be nice to see merged in its own PR - allowing non-local and local transforms to be used with `Combine` - this would actually probably fit in well with the current PR as well

Finally, I was also wondering if these changes will eventually end up being migrated to coreforecast instead, but I would say it'd be great to start merging in this repo and port them later once stable

nasaul:
Okay, I agree that it should be another PR. Going into `coreforecast` can be a great option, but then we need to implement this functionality in C++, but let's first see if the implementation is stable. I'll work on the docs and the refactor into Narwhals.

nasaul (approved):
Thanks for this PR, great work!

PR #651, opened 2026-05-20
Adding `partition_by` to lag-transforms building upon PooledState

simonez-tuidi:
## Summary

Depends on #641. This branch is stacked on `feature/groupby_with_range_semantics`, so #641 should be merged first. Until then, this PR includes that branch’s range-semantics and aggregate fast-path work in its history.

Fixes #650.

This PR adds `partition_by` support to pooled lag transforms across local, global, and groupby scopes. It routes pooled transforms through a unified state registry, tracks parent-calendar ordinals for partitioned buckets, and supports dynamic or unseen partition values during prediction and update.

## Prediction Validation

The latest commit updates `predict` validation to align with `drop_auxiliary_columns` semantics:

- `X_df=None` now raises when dynamic exogenous or `partition_by` columns are required.
- Static columns provided in `X_df` now warn and are ignored instead of raising.
- Missing required dynamic exogenous or `partition_by` columns still raise a hard error with a clearer message.

## Test Coverage

Adds or updates coverage for:

- partitioned fit, predict, update, dynamic keys, unseen buckets, datetime updates, sibling calendar advancement, backup/restore, and validation errors
- pandas and polars paths
- SQLite oracle checks for partitioned rolling and expanding pooled transforms
- EWM partition semantics with gapped parent ordinals
- updated `X_df` validation behavior for #650

PR #658, opened 2026-06-04
[FIX] Treat null/NaN keys as equal in pooled transforms

simonez-tuidi:
## Summary

Pooled `global_`/`groupby` lag transforms bucket rows by their group keys using joins. Those joins split **null**, **NaN**, and **pandas-`None`** into separate buckets (or dropped them entirely) instead of collapsing all missing values into a single bucket. As a result, a series with a missing groupby key could fan out across multiple buckets or fail to match between fit and predict — and could spuriously trip the "static feature changes over time" guard.

This PR makes missing-key handling consistent and null-safe across pandas and polars and across all key dtypes.

## Changes

- **Sentinel encoding for keys.** Every key column is encoded to a NUL-prefixed sentinel string before bucketing/joining, so *missing matches missing — and only missing*. This replaces the previous `match_if_categorical` reconcile path.
- **Shared null-equal join.** `add_bucket_id`, `lookup_bucket_ids`, `_attach_bucket_id`, and `_extend_groups` now route through one helper (`_null_equal_left_join`) so bucket creation, lookup, and extension all agree on what counts as the same key. New buckets in `_extend_groups` are created on the encoded key, and existing missing buckets are re-looked-up rather than duplicated.
- **Narrow int↔float reconcile.** When one side's key is integer and the other float (e.g. a column contaminated to float by a NaN, matched against clean integer data), integral float values are cast to int so `0.0` matches `0`. There is **no** blanket Float64 widening, so large integer keys above `2**53` stay distinct, and a genuinely fractional float (`1.5`) keeps its own bucket.
- **Null-safe static-feature check** (`mlforecast/core.py`). The "declared static but changes over time" check now treats two missing values as equal, so an all-null static feature (e.g. a null groupby key) is no longer flagged as changing. Backend-agnostic via narwhals.

## Tests

Adds coverage in `tests/test_pooled.py` and `tests/test_pooled_sqlite_oracle.py`:

- Missing-value collapse: null/NaN/None all land in one bucket; cross-engine (pandas/polars) parity.
- Int/float edge cases: mixed int/float lookup, fractional floats don't collide with int buckets, large int keys stay distinct, the int side is not widened by reconcile.
- `_extend_groups` doesn't duplicate an existing missing bucket; a partially-missing multi-column key keeps distinct buckets.
- End-to-end `groupby` with a null key through fit → predict → update, and no spurious static-change error.
- SQLite oracle parity tests for groupby null keys (string and numeric) and multi-column groupby with one null column.

## Notes / scope

- Scope is `global_`/`groupby` pooled transforms. The `partition_by` half of null-key handling is **not** included here and remains follow-up work.


simonez-tuidi:
@nasaul do you think you can review and merge this and then I can incorporate this fix into #651 as well so we can close all my open PRs?

Thanks in advance for all your help!

nasaul (approved):
Great work! We're grateful for contributions like this. Thanks for helping improve the project!

PR #666, opened 2026-06-17
tests: run distributed tests on linux only

jmoralez:
## Description

The distributed interface is meant only for Linux since it should only be used when there are several machines setup (otherwise the local will always be faster/more efficient). This skips the distributed tests in macos since they take a considerable amount of time ([22 min on macos](https://github.com/Nixtla/mlforecast/actions/runs/27652824904/job/81780667963?pr=665) vs [8 min on linux](https://github.com/Nixtla/mlforecast/actions/runs/27652824904/job/81780667988?pr=665)) reducing the org allocation for those resources and does not represent a supported use case.

Checklist:
- [ ] The tests pass.
- [ ] All linting tasks pass.

### DIFF
--- a/tests/test_distributed_forecast.py
+++ b/tests/test_distributed_forecast.py
@@ -12,6 +12,9 @@ from mlforecast.distributed.models.dask.lgb import DaskLGBMForecast
 from mlforecast.lag_transforms import ExpandingMean, RollingMean
 from mlforecast.utils import generate_daily_series
 
+if sys.platform != "linux":
+    pytest.skip("Distributed interface is only supported on Linux", allow_module_level=True)
+
 warnings.simplefilter("ignore", FutureWarning)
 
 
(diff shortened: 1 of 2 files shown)


tests/test_distributed_forecast.py
```diff
@@ -69,8 +72,6 @@ def fit(self, X, y, sample_weight=None):  # noqa: ARG002, D401, N803
         self.model_ = _RecordingLocalModel(weights)
         return self
 
-@pytest.mark.skipif(sys.platform == "win32", reason="Distributed tests are not supported on Windows")
-@pytest.mark.skipif(sys.version_info <= (3, 9), reason="Distributed tests are not supported on Python < 3.10")
```
jmoralez, line 73: the CI only runs for >= 3.10 so this check isn't needed anymore

nasaul (approved):
LGTM

PR #671, opened 2026-06-19
core: support non-jitted transformations

### LINKED ISSUE #494: [Feature Extraction] Allow using non numba-based custom features
### Description

It would be nice being able to use custom lag transformations that do not rely on numba.
For example, I was exploring the usage of the [catch 22](https://github.com/DynamicsAndNeuralSystems/pycatch22) features. The implementation would be straightforward:

```python
import pycatch22
import numpy as np

def rolling_co_trev_1_num(x: np.ndarray, window_size: int) -> np.ndarray:
    n = len(x)
    result = np.empty(n)
    result[:] = np.nan
    for i in range(window_size - 1, n):
        window = x[i - window_size + 1 : i + 1]
        feat = pycatch22.CO_trev_1_num(window.tolist())
        result[i] = feat
    return result

if __name__ == "__main__":
    print(rolling_co_trev_1_num(np.random.randint(10, 100, 100), 20))
```

However, I cannot use this function with the `MLForecast` class since numba cannot compile it (due to the usage of the `pycatch22` module inside the function).

### Use case

Expand the range of allowed lag transformations in the `MLForecast` class, thus opening the path to (hopefully) more accurate machine learning forecasting models.

jmoralez:
## Description

Fixes #494

Checklist:
- [ ] The tests pass.
- [ ] All linting tasks pass.

### DIFF
--- a/mlforecast/grouped_array.py
+++ b/mlforecast/grouped_array.py
@@ -2,6 +2,7 @@ __all__ = ['GroupedArray']
 
 
 import concurrent.futures
+import warnings
 from typing import Any, Dict, Mapping, Tuple, Union
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


nasaul (approved):
Great job @jmoralez thanks for your contribution

PR #678, opened 2026-06-29
# Pooled lag transforms: drop dead `ga` state + cheaper `predict` backup

simonez-tuidi:
## Summary

Two independent, **behavior-preserving** improvements to the pooled lag-transform state (`PooledState`, used by `global_=` / `groupby=` / `partition_by=` transforms), building on the merged `partition_by` work (#651):

1. **Remove dead state** — the `GroupedArray` mirror (`ga`), the pooled warm-up pass that was its only reader, and the write-only `_TimestampAggregates` fields `n_rows` / `is_balanced`.
2. **Cheaper `predict` backup** — replace the per-model `deepcopy` of every pooled state with a structural snapshot/restore that copies references, not array data.

No public API or behavior changes. This is the first of a short series of pooled-state follow-ups; pooled-state trimming under `keep_last_n` is a separate PR that builds on this one.

## Motivation

These are pre-existing inefficiencies in `PooledState`:

- **`ga` is dead weight.** `PooledState.ga` was kept in sync on every append (`append_several`) and rebuilt in each constructor, but its *only* reader was the pooled branch of `_initialize_lag_transform_states`, which called `state.ga.apply_transforms(...)` and **discarded the result**. All real feature computation goes through the flat arrays + `_ts_aggs` via `compute_pooled_features`. So `ga` cost memory in every state, a discarded `apply_transforms` pass at fit, and an `append_several` per recursive predict step — for nothing.
- **`n_rows` / `is_balanced` are write-only.** Computed in `_build_ts_aggs` and maintained on every append, never read anywhere (grep-verified).
- **`_backup` deep-copies everything per model.** Recursive `predict` wraps each model in `_backup`, which did `copy.deepcopy(self._pooled_states)` — i.e. it duplicated every aggregate and flat array of every pooled state, once per model per `predict`/`cross_validation` call. On large panels (many series, long history) that is a large, repeated time + peak-memory cost.

### 1. Remove dead `ga` mirror + warm-up + write-only fields

Dropped the `ga` field, its per-constructor `GroupedArray`/`process_df` construction, and every `append_several` sync site. Removed the pooled loop in `_initialize_lag_transform_states` (the local coreforecast warm-up is unchanged — those transforms genuinely need it). Removed `n_rows` / `is_balanced` from `_TimestampAggregates` and all their build/append/dynamic-bucket sites.

**Deliberately kept** (these only *looked* like `ga` machinery — they feed live paths): the `ufp.sort` orderings that keep `bucket_df` and the flat arrays mutually aligned; the `ga_data_dtype` → `float` cast chain on `y` (preserves the model's working-dtype rounding, e.g. float32, bit-for-bit); and `group_uids` → `match_if_categorical` → `new_bid` in grouped `append_observations` (drives bucket-id assignment, not just `ga`).

### 2. Snapshot/restore `_backup`

Added `PooledState.snapshot()` / `restore()`. Recursive prediction only ever **replaces** the array fields wholesale (`np.append` / `np.concatenate` return new arrays) and rebinds attributes on the `_TimestampAggregates` instances — it never mutates a shared array in place. So a faithful backup needs only reference copies of the arrays plus shallow copies of the mutable containers and of each aggregate instance; **no array data is duplicated**.

`_backup` now snapshots each pooled state and restores it in place on exit (the `ga` shallow-copy and `transforms` deepcopy are unchanged). The snapshot covers every field mutated during prediction: the flat arrays, `series_bucket_id`, `groups`, `bucket_df`, `_idsorted_to_bucket_pos`, `next_time_index_by_bucket`, the parent maps/grids, and `_ts_aggs`.

## Testing

All correctness rests on guards added in `tests/test_pooled_state_cleanup.py`:

- **G1 — byte-identical predictions.** A deterministic fit spanning all five pooled modes (global / groupby / local-partition / global+partition / groupby+partition) and every aggregate field (mean/std/min/max/expanding/EWM), with **two** models (so `_backup` runs between them) and multi-horizon `predict`, pinned to golden predictions captured on the pre-change baseline. This is the proof that the cleanup changes no output.
- **`snapshot` ≡ `deepcopy`.** After a full multi-model `predict`, every pooled state must be field-for-field equal to a `deepcopy` taken before predict; plus a focused case that restores correctly after `update_series_bucket_id` creates a brand-new dynamic bucket. This guards against any field the cheap snapshot might miss.

Full suites pass: `tests/test_pooled.py`, `tests/test_pooled_sqlite_oracle.py`, `tests/test_core.py`, `tests/test_forecast.py`, `tests/test_pooled_state_cleanup.py` — **701 passed, 2 skipped**. (`tests/test_distributed_forecast.py` requires dask, not run locally.)

## Risk / compatibility

Internal-only; no public API or behavior change. Two existing tests that asserted on the removed `ga` mirror were updated to the surviving representation (`bucket_id` / `y` per-bucket multiset; fixture no longer passes `ga=`).

## Note for reviewers

The functional change to `tests/test_pooled.py` is a single fixture (~5 lines). That file is **not** ruff-formatted on `main`, so letting the pre-commit `ruff-format` hook touch it would add ~1,100 lines of unrelated reformatting; the fixture commit therefore intentionally skips the format hook for that one file. Reformatting `test_pooled.py` is better done as its own dedicated `style:` change.

## Commits

1. `test: G1 byte-identical pooled-state guard across all modes`
2. `pooled: remove dead ga mirror, warm-up loop, and write-only agg fields`
3. `core: snapshot/restore _backup instead of deepcopy of pooled states`


nasaul (approved):
Thanks for this great work that you've been adding into the library!

PR #679, opened 2026-06-30
[FIX] Trim pooled lag-transform states under `keep_last_n`

simonez-tuidi:
# Trim pooled lag-transform states under `keep_last_n`  

## Summary  

`keep_last_n` only ever trimmed `TimeSeries.ga` (the per-series arrays used by local transforms); the **pooled** lag-transform states (`global_` / `groupby` / `partition_by`) kept the *entire* training history regardless. This PR brings pooled states to parity with `ga`: a pooled state whose transforms are all **finite-window** is trimmed to the minimum history it needs, while a state containing any **unbounded** transform (`Expanding*` / `ExponentiallyWeightedMean`) keeps full history.  

The win is memory: on large panels every pooled state previously grew to the full `O(unique_timestamps)` and was deep-copied per model on every `predict` call. Finite-window states now retain only `max(keep_last_n, widest_window)` ordinals.  

> **Stacked on #678** (`feature/pooled-state-cleanup` — dead-state cleanup + > cheaper `_backup`). This branch is cut from #678's branch; please review/merge > #678 first, then this rebases cleanly onto `main`. The base for this PR should > be #678's branch so the diff shows only the trimming changes.  

## ⚠️ Behavior change (please read)  

Trimming uses the **resolved** `keep_last_n` — i.e. the value the user passed **or** the value inferred from the transforms (`max(update_samples)`), exactly as the existing `ga` trim does. As a result:  

**`fit(df)` / `preprocess(df)` with no explicit `keep_last_n` now trims finite-window pooled states by default** (to the inferred widest window).  

This is intentional (parity with `ga`, and it delivers the memory win without the user having to opt in). It is **prediction-neutral** — the dropped prefix can never enter a finite window — but it does change the *internal representation* of pooled states. 13 existing tests that inspected the full pooled-state representation (calendars / ordinals / `bucket_df` length) were updated to pass an explicit large `keep_last_n`, so they keep exercising the full-history machinery; the trimming contract itself is owned by the new G2 guards.  

If preferred, restricting trimming to an *explicit* `keep_last_n` (leaving the inferred case untrimmed, so no existing tests change) is a one-line change — happy to switch.  

## Design  

**Which states may be trimmed.** A state is trimmable iff **every** transform assigned to it is finite-window. Detected with a new `_is_finite_window` property on `_BaseLagTransform`:  

- `True` for `Lag`, `Rolling*`, `SeasonalRolling*`. 
- `False` for `Expanding*` and `ExponentiallyWeightedMean` — pooled   expanding/EWM carry **no running accumulator** (unlike the local coreforecast   path); they recompute over the full aggregate vectors at every predict step,   so trimming them would change predictions. 
- `Offset` / `Combine` delegate to their wrapped transform(s). 
- Base default is `False`, so an unknown/custom transform is never silently   trimmed (correctness over the perf win).  

**Retention rule.** A trimmable state keeps its last `R = max(keep_last_n, W_state)` parent-calendar ordinals, where `W_state` is the state's widest window (`max(update_samples)`). The `W_state` floor is **required** and is where pooled legitimately diverges from local: local rolling survives an undersized explicit `keep_last_n` because coreforecast keeps a per-transform window buffer; pooled has none (the per-timestamp aggregates *are* the buffer), so a `keep_last_n` below a pooled window is raised to the window for that state. When `keep_last_n` is inferred it already equals the widest window, so the floor is then a no-op.  

**Implementation.** `PooledState.trim_to_last(n)` is defined as "fit-on-the-last-n-ordinal-slice" and built that way: it masks + renumbers the flat arrays and `bucket_df` (row-aligned at fit time), trims the parent grids, then **regenerates** the derived structures (`_ts_aggs`, `_idsorted_to_bucket_pos`) through the same primitives the constructors/append paths use — so every representation stays mutually consistent by construction rather than by hand-proven coordinate math. In all modes `next_time_index_by_bucket[bid]` is the bucket's calendar length, so the cutoff `len − n` is uniform across buckets that share a calendar, and renumbering (subtract cutoff) equals a fresh `searchsorted` into the retained suffix — which is exactly what the next `append_observations` recomputes. The caller (`TimeSeries._trim_pooled_states`, hooked right after the `ga` trim in `_transform`, fit-time only) computes the trimmable flag and `R`.  

## Correctness / validation  

New guards in `tests/test_pooled_keep_last_n_trim.py`:  

- **G2.1 prediction-equality** — trimmed vs untrimmed predictions match across   all five modes, including `keep_last_n` below the window (exercises the   retention floor). Compared with `rtol/atol=1e-9`, not byte-for-byte: the   rolling fast path cumsums the *whole* aggregate vector, so a shorter (trimmed)   vector accumulates window sums through different-magnitude partials (~1e-13   float-associativity noise, not a regression). 
- **G2.2 trim ≡ fit-on-slice** — a trimmed state is byte-identical (flat arrays,   `_ts_aggs`, parent grids, calendars, `_idsorted_to_bucket_pos`) to a fresh   state fit on only the retained tail, **before and after a follow-up   `update()`** (exercises the rebuilt aggregates and both append conventions).   Runs under **pandas and polars**. 
- **G2.3 suffix invariant** — each retained aggregate vector equals the   untrimmed tail, and retained length `== max(keep_last_n, W_state)`. 
- **G2.4 non-trim** — `Expanding*`/`EWM` states (and a mixed / `Combine`-with-   unbounded state) keep full history; `Offset`/`Combine` over finite transforms   stay trimmable.  

The byte-identical guarantee lives at the state level (G2.2/G2.3). The existing **SQLite-oracle** suite also runs with trimming active across every mode and both engines, confirming end-to-end that trimming does not move predictions.  

**Full relevant suite green: 717 passed, 2 skipped** (`test_pooled`, `test_pooled_sqlite_oracle`, `test_pooled_state_cleanup`, `test_pooled_keep_last_n_trim`, `test_core`, `test_forecast`). Ruff clean.  

## Files

| File | Change |
|------|--------|
| `mlforecast/lag_transforms.py` | `_is_finite_window` property (+ delegations) |
| `mlforecast/pooled.py` | `PooledState.trim_to_last(n)` |
| `mlforecast/core.py` | `_trim_pooled_states()` + hook in `_transform` |
| `tests/test_pooled_keep_last_n_trim.py` | new G2 guards |
| `tests/test_pooled.py`, `tests/test_core.py` | explicit `keep_last_n` on full-history-representation tests |
| `nbs/docs/how-to-guides/pooled_lag_transforms.ipynb` | "`keep_last_n` and pooled history" section + Memory bullet |

## Commits  

1. `pooled: trim pooled states under keep_last_n (finite-window states only)` — implementation + existing-test fixups
2. `test: G2 trim-correctness guards for keep_last_n pooled trimming`
3. `docs: pooled guide -- keep_last_n trimming behavior + divergence from local`

(Commit 1 and the docs commit used `--no-verify`: `test_pooled.py` is not ruff-formatted on this branch, and ruff-format's notebook hook rewrites the quote style of every code cell — both unrelated churn. Source files were ruff-checked/formatted and mypy-checked manually; this PR adds no new mypy errors.)  

## Out of scope (future PRs)  

- A carried accumulator for pooled `Expanding*`/EWM so even unbounded states can   trim. 
- Amortized array growth + cumsum caching; sliding-window fast paths for   `RollingQuantile`/`SeasonalRolling*`.

### DIFF
--- a/mlforecast/core.py
+++ b/mlforecast/core.py
@@ -342,6 +342,36 @@ class TimeSeries:
             local[name] = tfm
         return local
 
+    def _trim_pooled_states(self) -> None:
+        """Trim each pooled state's history under ``keep_last_n``.
+
+        Parity with the ``self.ga`` trim: a pooled state whose transforms are
+        *all* finite-window drops its unused history prefix, while a state
+        containing any Expanding*/EWM transform keeps full history (pooled has
+        no carried accumulator -- it recomputes over the full aggregate vectors
+        at predict, so trimming those would move predictions).
+
+        Retention is ``max(keep_last_n, W_state)`` ordinals, where ``W_state``
+        is the state's largest finite window. The floor is required and is where
+        pooled legitimately diverges from the local coreforecast path: local
+        rolling survives an undersized explicit ``keep_last_n`` because
+        coreforecast carries a per-transform window buffer; pooled has none (the
+        aggregates *are* the buffer), so trimming below ``W_state`` would compute
+        windows off a truncated prefix. When ``keep_last_n`` is inferred it
+        already equals the global max window, so the floor is then a no-op.
+        """
+        if self.keep_last_n is None:
+            return
+        for key, tfms in self._get_pooled_tfms().items():
+            state = self._pooled_states.get(key)
+            if state is None:
+                continue
+            tfm_list = list(tfms.values())
+            if not all(tfm._is_finite_window for tfm in tfm_list):
+                continue
+            w_state = max(tfm.update_samples for tfm in tfm_list)
+            state.trim_to_last(max(self.keep_last_n, w_state))
+
     def _initialize_lag_transform_states(self) -> None:
         """Materialize lag transform state for subsequent update-based prediction.
 
@@ -910,6 +940,7 @@ class TimeSeries:
             self.keep_last_n = max(update_samples)
         if self.keep_last_n is not None:
             self.ga = self.ga.take_from_groups(slice(-self.keep_last_n, None))
+            self._trim_pooled_states()
         del self._restore_idxs, self._sort_idxs
 
         # lag transforms
--- a/mlforecast/core.py
+++ b/mlforecast/core.py
@@ -346,18 +346,15 @@ class TimeSeries:
         """Materialize lag transform state for subsequent update-based prediction.
 
         This is needed when a new ``TimeSeries`` instance is created from historical
-        data right before calling ``predict(new_df=...)``. Local, global and grouped
-        transforms all need to see a full ``transform`` pass so stateful transforms
-        like ``ExpandingMean`` can initialize their internal buffers before the
-        first ``update(...)`` call.
+        data right before calling ``predict(new_df=...)``. Local (coreforecast)
+        transforms need a full ``transform`` pass so stateful transforms like
+        ``ExpandingMean`` can initialize their internal buffers before the first
+        ``update(...)`` call. Pooled transforms keep their state in ``_ts_aggs``
+        (built at construction), so they need no warm-up pass.
         """
         core_tfms = self._get_core_lag_tfms()
         if core_tfms:
             self._compute_transforms(core_tfms, updates_only=False)
-        pooled_tfms = self._get_pooled_tfms()
-        for key, tfms in pooled_tfms.items():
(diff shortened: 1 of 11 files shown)


nasaul (approved):
Thanks for your contribution!

PR #672, opened 2026-06-19
pre-commit: add ruff-format

jmoralez:
## Description

* Bumps ruff-pre-commit's version
* Runs both the linter and formatter
* Applies the formatter changes to have a consistent style across the codebase.

All the changes in this PR were automatic by running the same command as the lint task (`uvx pre-commit run --files mlforecast/*`) 

Checklist:
- [ ] This PR has a meaningful title and a clear description.
- [ ] The tests pass.
- [ ] All linting tasks pass.
- [ ] The notebooks are clean.

PR #682, opened 2026-07-02
[FIX] default `min_samples` to 1 in local partition mode (`partition_by`)

simonez-tuidi:
## Description

Follow-up to #651. Fixes the `min_samples` default for local partition mode (`partition_by` without `global_`/`groupby`), where inheriting the local-mode default of `window_size` makes the feature structurally all-NaN.

### Problem

`min_samples=None` resolves to `window_size` everywhere. In plain local mode that's a well-defined "full window only" guard: windows are row-based per series, so `window_size` and the sample count are the same unit. Under `partition_by` they are not — the RANGE window spans `window_size` **parent-calendar steps**, while `min_samples` counts **same-partition observations** inside it. In local partition mode the guard can only pass when a series has `window_size` consecutive same-partition steps:

```python
# dense daily panel, promo interleaved (~30% of days)
RollingMean(7, partition_by=["promo"])                  # ~100% NaN
RollingMean(7, min_samples=1, partition_by=["promo"])   # NaN share ~1-4%
```

A 7-calendar-day window almost never holds 7 same-promo observations, so the default silently produces an unusable feature. `min_samples=1` matches the SQL `RANGE BETWEEN ... PRECEDING` semantics the pooled transforms are documented (and oracle-tested) against: SQL window aggregates return NULL only for empty frames and have no minimum-count guard.

### Change

`min_samples=None` now resolves per mode:

| Mode | Default |
| --- | --- |
| local (per series, coreforecast) | `window_size` (unchanged) |
| `global_` / `groupby` | `window_size` (unchanged) |
| `global_`/`groupby` + `partition_by` | `window_size` (unchanged, counted across all series in the (group, partition) bucket) |
| **`partition_by` alone (local partition)** | **1** |

- A single `_resolve_min_samples()` helper replaces the 15 inline `self.min_samples if ... else window_size` resolutions across the rolling and seasonal-rolling classes, so fit, the aggregate fast paths, and the predict-time `_compute_latest_from_aggs` path all agree.
- Explicit `min_samples` values are never overridden.
- The `min_samples` docstrings (rolling + seasonal) and the pooled lag transforms guide now document the mode-dependent default, including the combined `global_`/`groupby` + `partition_by` behavior.

This changes default behavior for local partition mode only. Since `partition_by` shipped in #651, no released default is affected beyond that feature, and any previous local-partition output under the old default was almost entirely NaN.

## Test Coverage

- The SQLite oracle (`assert_oracle_matches`) derives the local-partition default **independently** of the library helper, so it pins the default rather than mirroring it. The existing local-partition oracle tests previously compared mostly-NaN against mostly-NULL under the old default; they now exercise real window values on both sides.
- New unit test for the resolution across all six mode combinations (rolling + seasonal, explicit override).
- New integration test reproducing the motivating scenario: dense panel with interleaved promo days, asserting the default equals `min_samples=1` end-to-end and that the old default (`min_samples=window_size`) yields all NaN.
- New test pinning that `global_` + `partition_by` keeps the `window_size` default.

## Commit structure

The first commit is a formatting-only `ruff format` pass over `tests/test_pooled.py` and `tests/test_pooled_sqlite_oracle.py`, which predate the ruff-format pre-commit hook added in #672 — kept separate so the semantic diff stays reviewable. The behavior change is entirely in the second commit.

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.


nasaul:
Hey @simonez-tuidi thanks for your contribution this is a great job, just fixing the main merging and there's an issue with coreforecast that I need to solve first before the testing suite passes No need to worry on your end

nasaul (approved):
Thanks for your contribution

PR #683, opened 2026-07-04
[FEAT] Add LookupLag for event-based target lookups

### LINKED ISSUE #675: [Feature request] Support a series lag grouped by a dynamic column
### Description

I have daily series with a `holiday_name` column that marks national holidays. For each holiday, I'd like a feature that holds last year's `y` for that same holiday.

The catch is moving holidays like Easter: they land on a different date every year, so a fixed lag won't find them. The right value has to be looked up per `(unique_id, holiday_name)`, and `holiday_name` changes from row to row (it's empty on non-holiday days).

I'd like this feature to be computed within the `MLForecast` class rather than computed by myself, so it stays **consistent** with any `target_transforms`, just like the built-in lag features.

The main problem is precisely `target_transforms`. If I precompute the feature externally, it sits on the raw `y`, while `MLForecast` trains on the transformed target and its lag features. For example, with `LocalStandardScaler` the target and the regular lags are standardized per series, but my precomputed `y_same_holiday_last_year` is not, so the feature and the target end up on different scales and aren't aligned. To fix it I'd have to replicate the exact per-series scaling myself, which is fragile.

### Reproducible example

```python
import numpy as np
import pandas as pd

np.random.seed(0)

# Moving holiday (easter) + fixed-date holiday (christmas) on a daily panel.
easter = {2021: "2021-04-04", 2022: "2022-04-17", 2023: "2023-04-09", 2024: "2024-03-31"}

rows = []
for uid in ["store_1", "store_2"]:
    dates = pd.date_range("2021-01-01", "2024-12-31", freq="D")
    d = pd.DataFrame({"unique_id": uid, "ds": dates, "y": np.random.rand(len(dates)).round(3)})
    d["holiday_name"] = pd.Series([None] * len(dates), dtype="object")
    for _, dt in easter.items():
        d.loc[d["ds"] == dt, "holiday_name"] = "easter"
    d.loc[d["ds"].dt.strftime("%m-%d") == "12-25", "holiday_name"] = "christmas"
    rows.append(d)
df = pd.concat(rows, ignore_index=True)

# Desired feature: y of the same holiday, previous year.
# Group by (unique_id, holiday_name) because moving holidays change date.
hol = df[df["holiday_name"].notna()].copy()
hol["y_same_holiday_last_year"] = (
    hol.sort_values("ds")
       .groupby(["unique_id", "holiday_name"])["y"]
       .shift(1)
)
df = df.merge(
    hol[["unique_id", "ds", "y_same_holiday_last_year"]],
    on=["unique_id", "ds"], how="left",
)
print(df.dropna(subset=["holiday_name"]).to_string(index=False))
```
Output (Easter moves Apr 4 -> Apr 17 -> Apr 9 -> Mar 31, yet each row maps to the prior year's Easter value):
```
unique_id         ds     y holiday_name  y_same_holiday_last_year
  store_1 2021-04-04 0.716       easter                       NaN
  store_1 2021-12-25 0.393    christmas                       NaN
  store_1 2022-04-17 0.943       easter                     0.716
  store_1 2022-12-25 0.896    christmas                     0.393
  store_1 2023-04-09 0.803       easter                     0.943
  store_1 2023-12-25 0.913    christmas                     0.896
  store_1 2024-03-31 0.367       easter                     0.803
  store_1 2024-12-25 0.272    christmas                     0.913
  store_2 2021-04-04 0.127       easter                       NaN
  store_2 2021-12-25 0.989    christmas                       NaN
  store_2 2022-04-17 0.382       easter                     0.127
  store_2 2022-12-25 0.849    christmas                     0.989
  store_2 2023-04-09 0.910       easter                     0.382
  store_2 2023-12-25 0.819    christmas                     0.849
  store_2 2024-03-31 0.541       easter                     0.910
  store_2 2024-12-25 0.025    christmas                     0.819
```

### Question

Is there a supported way to do this today that I've missed?

### Use case

_No response_

janrth:
## Summary

Adds `LookupLag`, a lag transform for occurrence-based target lookups within dynamic partition keys. This supports cases like holiday features where the relevant previous target is not a fixed calendar lag, for example Easter moving between dates year to year.

Solves #675

The lag value still comes from the existing `lag_transforms` dictionary key:

```python
from mlforecast.lag_transforms import LookupLag

fcst = MLForecast(
    models=[model],
    freq="D",
    lag_transforms={
        1: [LookupLag(partition_by=["holiday_name"])]
    },
    target_transforms=[LocalStandardScaler()],
)
```

This returns the previous transformed target observed within each `(unique_id, holiday_name)` bucket. Since it is computed inside MLForecast, it stays aligned with target transforms such as `LocalStandardScaler`.

## Why

Users can precompute features like `y_same_holiday_last_year` externally, but that happens before MLForecast target transforms. With local transforms such as `LocalStandardScaler`, external features stay on the raw target scale while model targets and built-in lag features are transformed. `LookupLag` keeps the lookup target-transform-aware.

## Examples enabled

Simple moving-holiday lookup:

```python
lag_transforms={
    1: [LookupLag(partition_by=["holiday_name"])]
}
```

This gets each store's previous value for the same holiday name, independent of the date shift.

A broader pooled signal can be combined with the individual event lookup:

```python
lag_transforms={
    1: [LookupLag(partition_by=["holiday_name"])],
    330: [
        RollingMean(
            window_size=120,
            min_samples=1,
            groupby=["region"],
            partition_by=["holiday_name"],
        )
    ],
}
```

Here `LookupLag` captures the individual store's previous same-holiday value, while the grouped partitioned rolling mean captures broader regional holiday behavior around last year.

## Implementation notes

- `LookupLag` subclasses `_BaseLagTransform` and plugs into `_compute_bucket_feature`, the existing pooled/partitioned transform hook.
- The existing `lag_transforms` key is interpreted as the occurrence lag for `LookupLag`.
- The transform reuses existing `partition_by` machinery, prediction-time `X_df` handling, and target-transform ordering.

## Validation

- `uv run ruff check mlforecast/lag_transforms.py tests/test_core.py`
- `uv run ruff format --check mlforecast/lag_transforms.py tests/test_core.py`
- Added tests for pandas and polars covering:
  - previous matching holiday occurrence lookup
  - lookup values computed on transformed targets with `LocalStandardScaler`


janrth:
@codex

nasaul (approved):
Thanks for your contribution

PR #685, opened 2026-07-06
[FEAT] Add time_agg support for pooled lag transforms

### LINKED ISSUE #684: [Core] `time_agg` — per-timestamp pre-aggregation for pooled lag transforms
### Description

Pooled lag transforms (`global_=True`, `groupby=[...]`, `partition_by=[...]`) currently treat every observation as an individual sample: a `RollingMean(window_size=7, groupby=["category"])` pools all rows of the category that fall inside the window and averages them with row weighting, matching a SQL `RANGE` window over per-row values. This request adds an opt-in `time_agg` parameter that first collapses all rows sharing a timestamp within the pooling bucket into a single value — one of `"sum"`, `"count"`, `"mean"`, `"min"`, `"max"` — and then applies the lag transform over that per-timestamp series. `time_agg=None` (the default) keeps current behavior exactly.

```python
# rolling mean of the category's DAILY TOTALS, not of the pooled individual rows
RollingMean(window_size=7, groupby=["category"], time_agg="sum")
```

### Use case

In retail/demand settings a natural cross-series feature is "the trajectory of the group's total": the 7-day rolling mean of a category's total daily sales, the peak daily total over the last month (`RollingMax(..., time_agg="sum")`), the volatility of daily totals (`RollingStd(..., time_agg="sum")`), or the number of active series per day (`time_agg="count"`). None of these are expressible today: the pooled transforms only offer row-weighted statistics, where a day with 50 observations weighs 50× a day with one. The manual workaround — aggregate the frame per (group, timestamp), compute features on the aggregate series, merge back — works for `preprocess` but breaks at `predict` time, because recursive forecasting must update these features step by step as predictions are appended; only transforms integrated with the internal pooled state get that for free.

### Option A — a `time_agg` parameter on the existing transforms

Add `time_agg: Optional[str] = None` to the pooled-capable transforms, with a closed set of aggregations (`sum`, `count`, `mean`, `min`, `max`). Lean API change; the perceived risk is implementation complexity and parameter-combination confusion.

### Option B — dedicated classes (`RollingMeanWithAgg`, `ExpandingMeanWithAgg`, `SeasonalRollingMeanWithAgg`, ...)

New transform classes taking `agg` plus the usual pooled-mode parameters, initially limited to the Mean-applying variants.

## Recommendation: Option A (`time_agg` parameter)

- **The implementation cost of the two options is identical, so the choice is pure API surface.** Pooled transforms already materialize per-timestamp aggregates internally (`_TimestampAggregates`: per-(bucket, timestamp) `sums`/`counts`/`sum_sq`/`mins`/`maxs`, `mlforecast/pooled.py:171-182`), and every fast-path helper (`_rolling_mean_from_agg`, the std/min/max/expanding/ewm variants in `mlforecast/lag_transforms.py`) consumes only that structure. Either API reduces to one small adapter that derives a unit-count aggregate view (`sums=v, counts=0/1, sum_sq=v², mins=maxs=v` where `v` is the chosen per-timestamp aggregate) and feeds it to the unchanged helpers — including the predict-time fast path (`_compute_latest_from_aggs`) and the recursive-update machinery (`append_predictions`). New classes would still have to implement the same three-hook contract, so they buy zero implementation savings.
- **Option B explodes combinatorially.** Once the feature is framed as "apply the transform to the per-timestamp aggregated series", every statistic is meaningful: max of daily sums is peak daily demand, std of daily sums is demand volatility, expanding max of daily counts is record participation. The first user request beyond Mean forces `{Rolling, Expanding, Seasonal, EWM} × {Mean, Std, Min, Max, Quantile}` duplicate classes, each repeating the `lag/window_size/min_samples/global_/groupby/partition_by` surface.
- **The concept already exists implicitly in the codebase.** Pooled `ExponentiallyWeightedMean` already consumes each timestamp's bucket-aggregate mean exactly once regardless of how many rows share the timestamp (pinned by `tests/test_pooled.py::test_global_partition_ewm_uses_timestamp_mean_once`) — i.e. it already behaves as `time_agg="mean"`. A parameter makes this existing behavior explicit and uniform across transforms instead of a per-transform quirk.
- **Discoverability and composition favor the parameter.** One documented parameter on familiar classes beats a parallel class family; `Offset` and `Combine` delegate to their inner transforms, so `time_agg` composes with them without any new class-specific interop, and `Combine` can even mix inners with different `time_agg` values (e.g. rolling mean of sums ÷ rolling mean of means).
- **The name `time_agg` (rather than `agg`) addresses the confusion risk directly:** it reads as "how to aggregate rows sharing a timestamp, before the transform", and cannot be mistaken for the transform's own statistic. Feature names disambiguate automatically via the existing changed-parameter suffix (e.g. `groupby_category_rolling_mean_lag7_time_aggsum`), and names of existing `time_agg=None` configurations remain byte-identical.

## Proposed semantics and guardrails

- `time_agg` requires `global_=True` or `groupby=[...]` (with or without `partition_by`); construction raises `ValueError` otherwise, because in local modes each (bucket, timestamp) holds a single row and the aggregation is an identity — rejecting avoids silent no-ops.
- Closed aggregation set `{"sum", "count", "mean", "min", "max"}` — exactly what the internal per-timestamp cache supports; anything else raises `ValueError`.
- With `time_agg` set, `min_samples` counts observed timestamps in the window (each timestamp contributes 0 or 1), not rows — documented alongside the existing pooled `min_samples` semantics.
- All-NaN timestamps count as unobserved for `sum`/`mean`/`min`/`max` (matching SQL aggregates over all-NULL groups → NULL), while `time_agg="count"` yields the non-NaN row count, including 0 — matching SQL `COUNT`.
- `RollingMean(time_agg="mean")` differs from `RollingMean()` on unbalanced buckets (unweighted mean-of-daily-means vs row-weighted pooled mean) — this is the one genuinely confusable pair and gets an explicit docs callout; on `ExponentiallyWeightedMean`, `time_agg="mean"` is accepted and documented as equivalent to the current pooled behavior.

## Scope and feasibility notes

- Coverage: all pooled-capable transforms. The nine with aggregate fast paths (Rolling/Expanding Mean/Std/Min/Max, EWM) go through the adapter over the cached aggregates; the row-level-only transforms (`RollingQuantile`, `SeasonalRolling*`, `ExpandingQuantile`) are covered by collapsing raw rows to one row per (bucket, timestamp) before the existing row loops — same semantics, one extra free function.
- Correctness is pinned independently of the implementation by extending the existing SQLite window oracle (`tests/test_pooled_sqlite_oracle.py`) with a per-timestamp `GROUP BY` CTE between the ordinal computation and the window aggregation, so the SQL engine defines the expected values for every (transform × agg × lag × mode) combination.
- No changes to mode resolution or state construction: `time_agg` is not part of the pooled-state key, so transforms differing only in `time_agg` share the same cached state; the adapter derives fresh non-mutating views per call, keeping multi-step recursive prediction consistent as the cache is updated in place.
- `preprocess` and `predict` both work unchanged from the user's perspective, including `partition_by` combinations and dynamic partition values via `X_df`.

simonez-tuidi:
This PR adds a `time_agg` parameter to the pooled lag transforms (`global_=True`, `groupby=[...]`, optionally combined with `partition_by`). When set to one of `"sum"`, `"count"`, `"mean"`, `"min"`, `"max"`, all rows sharing a timestamp within a bucket are first collapsed into a single per-timestamp value, and the transform is then applied over that per-timestamp series — e.g. `RollingMean(window_size=7, groupby=['category'], time_agg='sum')` computes a rolling mean of the category's **daily sums** instead of the row-weighted pooled mean. The default `time_agg=None` keeps the current behavior exactly (every observation is an individual sample).

This makes features like "trajectory of the group's total" first-class: 7-day rolling mean of a category's total daily sales, peak daily total over the last month (`RollingMax(..., time_agg='sum')`), volatility of daily totals (`RollingStd(..., time_agg='sum')`), or the number of active series per day (`time_agg='count'`). These were previously inexpressible — the manual workaround (aggregate the frame, compute features, merge back) breaks at `predict` time because recursive forecasting must update the features step by step.

Closes #684 

### API and semantics

- `time_agg` is accepted by all pooled-capable transforms: `Rolling{Mean,Std,Min,Max,Quantile}`, `SeasonalRolling*`, `Expanding{Mean,Std,Min,Max,Quantile}`, `ExponentiallyWeightedMean`. `Offset` and `Combine` delegate to their wrapped transforms (`Combine` may mix inners with different `time_agg` values).
- Requires `global_=True` or `groupby=[...]`; construction raises `ValueError` otherwise, since local and `partition_by`-only modes have a single row per (bucket, timestamp) and the aggregation would be an identity.
- **`ExponentiallyWeightedMean` is the one exception**: its `time_agg` defaults to `"mean"` (and `None` is rejected), because pooled EWM already consumes each timestamp's bucket-aggregate mean exactly once — the default codifies the pre-existing behavior, and values/feature names are unchanged for existing configurations.
- With `time_agg`, `min_samples` counts observed **timestamps** in the window (each contributes 0/1), not rows — documented alongside the existing pooled `min_samples` semantics.
- All-NaN timestamps count as unobserved for `sum`/`mean`/`min`/`max` (matching SQL aggregates over all-NULL groups → NULL); `time_agg='count'` yields the non-NaN row count including 0 (matching SQL `COUNT`).
- Feature names gain an automatic suffix only when `time_agg` differs from the default (e.g. `groupby_category_rolling_mean_lag7_window_size7_time_aggsum`), so all existing feature names are byte-identical.

### Implementation

- **Template-method hooks on `_BaseLagTransform`.** The public pooled hooks (`_compute_bucket_feature`, `_compute_latest_from_aggs`, `_compute_ts_level_from_aggs`) live on the base class: they guard empty inputs, apply the `time_agg` re-aggregation exactly once, and dispatch to `*_impl` methods that return `None` on the base when a path is unsupported. Subclasses only implement the `_impl` methods and never touch `time_agg` themselves, so a transform that gains a new fast path cannot silently skip re-aggregation. `Offset`/`Combine` override the public hooks and delegate to their wrapped transforms, so each inner keeps its own `time_agg` (`Combine` may mix inners with different values).
- **One value derivation, shared by both paths.** A single `_time_agg_values(agg, time_agg)` helper is the sole source of the `sum`/`count`/`mean`/`min`/`max` value logic (including the all-NaN → NaN vs `count` → 0 conventions), used by both the aggregate fast path and the row slow path so they cannot diverge.
- **Lazy re-aggregation view.** `_reaggregate_ts_aggs` returns `_ReaggregatedAggregates` views over the cached per-timestamp aggregates, exposing a unit-count shape (`sums=v, counts=0/1, sum_sq=v², mins=maxs=v`) to the existing `_*_from_agg` helpers unchanged — including the predict fast path and recursive-update machinery. Everything (the collapsed value array itself and each of the five aggregate fields) is derived on first access, so re-aggregating before an unsupported hook that returns `None` does no numpy work. The views never mutate the shared `PooledState._ts_aggs` cache, preserve `unique_times` exactly (the fit-path dense-position mapping depends on it), and are recomputed per call so they can never go stale after `append_predictions`.
- **Slow-path transforms reuse the cache.** Row-level transforms (quantiles, seasonal) go through `_collapse_rows_by_time`, which collapses rows to one per (bucket, timestamp) before the existing row loops. When the per-timestamp aggregate cache is available it derives the collapse from it in O(unique timestamps) instead of re-aggregating every raw row, and returns a row→collapsed inverse-index array so the per-timestamp results broadcast back with a single gather. Fast and slow paths are pinned equal by a property test: `_build_ts_aggs(collapse(rows)) == _reaggregate_ts_aggs(_build_ts_aggs(rows))`.
- **EWM `"mean"` is a no-op, not a copy.** `ExponentiallyWeightedMean` overrides `_pooled_time_agg` to return `None` for `time_agg="mean"`, so the default configuration short-circuits re-aggregation entirely (its update rule already consumes each timestamp's bucket mean once) — bit-identical output, zero per-call allocation.
- **Validation and `Offset` hygiene.** A single `_validate_time_agg(..., allow_none=, scope_exempt=)` covers both the general policy and EWM's (`None` rejected, `"mean"` scope-exempt). `Offset` no longer mirrors `time_agg` onto the wrapper (nothing reads it there — the delegated inner applies its own), the base class carries `time_agg = None` for uniform introspection, and `Offset._set_core_tfm` rejects an effective lag `< 1`.
- `time_agg` is not part of the pooled-state key: transforms differing only in `time_agg` share one `PooledState`. No changes to `core.py`.

### Testing

- SQLite window oracle extended with an optional per-timestamp `GROUP BY` CTE assembled into the shared CTE list between the ordinal computation and the window aggregation (the ordinal/window definitions are reused verbatim by both variants), with the final SELECT broadcasting per-timestamp results back to every original row via a LEFT JOIN — SQLite independently pins the semantics across the transform × agg × lag × global/groupby matrix plus the partition_by matrices (`tests/test_pooled_sqlite_oracle.py`, ~510 cases).
- `tests/test_pooled.py`: adapter unit tests (no-mutation, `unique_times` preservation, all-NaN conventions), the fast-vs-slow property test now also asserting the collapse's row→collapsed inverse-index mapping and cache-fed equivalence, hand-computed literal tests on unbalanced buckets, fast-vs-slow cache-wipe equivalence across all 9 fast-path transforms × 5 aggs, multi-step recursive predict, validation/warning tests, EWM default-equivalence and `"mean"` short-circuit tests, structural tests (feature names, `Offset` delegation, `Offset` effective-lag validation, `Combine`, sklearn `clone`/`get_params` round-trip), quantile/seasonal slow-path literals.
- Full suite passes locally (pandas + polars), including the M4 performance pins; `mypy` and `ruff` (check + format) clean via pre-commit.

### Docs

- New "Aggregating rows per timestamp with `time_agg`" section in the pooled lag transforms guide (`nbs/docs/how-to-guides/pooled_lag_transforms.ipynb`), plus an entry in its constraints list.
- Google-style docstring entries on every `__init__` (auto-rendered API reference).

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.
- [x] The notebooks are clean.ge.


simonez-tuidi:
@janrth this should do the trick, let me know what you think
@nasaul this one and #679 + #682 are next up in the list of review that I would need from you, sorry there's been so much to assess lately 😩

nasaul (approved):
Thanks for your contribution!

PR #688, opened 2026-07-10
[CHORE] Narwhals Refactor

### LINKED ISSUE #577: Narwhals implementation
### Description

Right now the library has two logics, one for `polars` and one for `pandas`. I think that the best way to go is to condense this into a single logic, this can be achieved with `Narwhals`.

### Use case

_No response_

nasaul:
## Summary

Refactors the codebase to use [Narwhals](https://narwhals-dev.github.io/narwhals/) as a single dataframe-agnostic path instead of maintaining parallel pandas/polars branches. This removes most of the hand-written backend splits across `core.py`, `forecast.py`, `data_validation.py`, and `optimization.py`, leaving native-backend code only where it's genuinely required (see below).

Closes #577

## What changed

- **`data_validation.py` / `optimization.py`** — unified pandas/polars branches through narwhals (schema-based categorical detection, shared validation logic). Object-dtype columns are still treated as categorical to preserve pre-refactor behavior.
- **`core.py`** — centralized `uids`/`last_dates` native conversion into one helper, rewrote op-splits (bucket-join construction, date features, residuals) via narwhals, and reset the pandas index after the narwhals horizon filter so output matches the pre-refactor result exactly.
- **`forecast.py`** — unified fitted-values horizon cast/filter via narwhals.

## Native-backend code that was intentionally kept

Not every split was incidental duplication. A handful are performance- or compatibility-motivated and remain backend-specific, now with plain-English comments explaining why (e.g. narwhals has no equivalent for `DatetimeIndex` attribute access; sklearn/CatBoost models consume pandas/numpy at the model boundary; the per-step null check stays vectorized in the recursive hot loop).

## Performance

The recursive `predict` hot loop is sensitive to per-call narwhals overhead. The per-step feature null-check was restructured so it runs on raw numpy feature values before frame assembly, statics are scanned once per `predict` in `_predict_setup`, and `X_df` is pre-scanned once — restoring the hot-loop cost to parity with `main` (local A/B: ~78 ms → ~30 ms).

## Testing

- `pre-commit` (ruff + mypy) passes on all changed files.
- CodSpeed benchmarks (`tests/test_pipeline.py::test_predict`, `test_preprocess`) back to base parity.

### DIFF
--- a/mlforecast/core.py
+++ b/mlforecast/core.py
@@ -235,6 +235,19 @@ def _static_feature_changes_over_time(start_series, end_series) -> bool:
     return bool(changed.fill_null(False).any())
 
 
+def _to_native_uids(values, *, df):
+    """Wrap ids/dates into the backend-native container mlforecast exposes as
+    ``TimeSeries.uids`` / ``TimeSeries.last_dates``.
+
+    pandas -> ``pd.Index``, polars -> ``pl.Series``. Centralized here because these
+    attributes are observable API (tests rely on ``.tolist()`` and fancy indexing),
+    so their native type must be produced in exactly one place.
+    """
+    if isinstance(df, pd.DataFrame):
+        return pd.Index(values)
+    return pl_Series(values)
+
+
 class TimeSeries:
     """Utility class for storing and transforming time series data."""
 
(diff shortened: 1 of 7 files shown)


mlforecast/core.py
```diff
@@ -1210,6 +1225,11 @@ def _update_y(self, new: np.ndarray) -> None:
 
     def _update_features(self) -> DataFrame:
         """Compute the current values of all the features using the latest values of the time series."""
+        features_df = self._compute_features_df()
```
JQGoh, line 1236: Reminder on this by Claude:

> _update_features() now silently depends on _predict_setup().
_compute_features_df() reads self._uniform_dates and writes self._feature_null_cols, both of which are only initialized in _predict_setup(). _uniform_dates is declared as a bare class annotation with no default, so calling _update_features() without a prior _predict_setup() raises AttributeError. All current callers (prod and tests) run _predict_setup() first, so this is not a live bug —but it's an implicit coupling that didn't exist before (the old _update_features was self-contained). Worth a one-line guard or a docstring note, since _update_features is called directly in several tests and is an easy future footgun.
```

JQGoh (approved):

PR #690, opened 2026-07-16
[FEAT] Add `history_warmup` to build the forecasting state without materializing features

simonez-tuidi:
This PR adds `MLForecast.history_warmup()` (backed by `TimeSeries.history_warmup()`), which ingests raw history and builds all the internal state required for forecasting — series storage, target-transform state, local (coreforecast) lag-transform buffers, pooled aggregate states, and the feature order — without computing or assembling the training features dataframe. It is the cheap alternative to `preprocess` for the case where the models are already trained and the features would be thrown away.

## Motivation

Today the public ways to get a predict-ready object are `fit` and `preprocess` — both of which pay the full cost of materializing every lag transform into a wide features dataframe — or the transfer-learning path `predict(new_df=...)`, which builds state cheaply but only works on an already-fitted instance and is fused to producing a forecast (see the comparison below). In serving scenarios — unpickling trained models, `MLForecast.load`, or assigning externally trained estimators to `models_` — that work is pure overhead: the user only wants to load history, optionally `update()` with fresh observations, and `predict()`. This also composes with the recent `keep_last_n` trimming work (#679): after warming up, history is stripped down to what updates actually need, with the transform accumulators kept in its place.

```python
fcst = MLForecast(models=[], freq="D", lags=[1], date_features=["dayofweek"])
fcst.models_ = trained_models  # unpickled or trained elsewhere
fcst.history_warmup(history_df)
fcst.update(new_observations)  # optional
fcst.predict(7)
```

## Relationship to `predict(new_df=...)`

The transfer-learning path `predict(h, new_df=history, X_df=future_exog)` already built the forecasting state from raw history without materializing features, and even persists the new state on `self.ts` after a successful call — for a *previously fitted* object, a one-shot forecast from new history was already covered. `history_warmup` is that internal mechanism lifted into a public method (and `predict(new_df=)` now delegates to it), with the following explicit differences:

- **Works on never-fitted objects.** `predict(new_df=)` reads its configuration (`id_col`/`time_col`/`target_col`, `static_features`, `keep_last_n`, `weight_col`, `max_horizon`, `_horizons`, `as_numpy`) from `self.ts` attributes that only exist after a `fit`/`preprocess`, so a fresh `MLForecast` with externally assigned `models_` fails with `AttributeError: 'TimeSeries' object has no attribute 'max_horizon'`. `history_warmup` takes these as arguments, so "construct + assign `models_` + warm up" now works without ever running `preprocess`.
- **Decoupled from forecasting.** State-building is no longer fused to the first predict: you can warm at load time, `update()` repeatedly as observations arrive, and predict on demand — including update-first workflows where no forecast precedes the first update.
- **Memory.** `predict(new_df=)` keeps the full provided history (behavior preserved via `trim=False`); `history_warmup` applies the `keep_last_n` trim like `fit`, so long histories collapse to what updates actually need.
- **Validation.** `predict(new_df=)` runs no data-quality checks on the provided history; `history_warmup` has the same `validate_data` prelude as `preprocess` (gap/duplicate detection, pooled gap-free warning).
- **Model-shape metadata.** For one-model-per-horizon setups, `history_warmup(max_horizon=...)`/`horizons=...` declares how external models were trained; `predict(new_df=)` can only inherit that from a previous fit.

## What changed

- `TimeSeries.history_warmup(df, ..., max_horizon=None, horizons=None, trim=True)`: runs `_fit` (validation, ga/uids/last_dates, target-transform `fit_transform`, static features, `features_order_`, pooled states), warms the local lag-transform buffers via `_initialize_lag_transform_states()` (required so stateful transforms like `ExpandingMean`/`ExponentiallyWeightedMean` can serve `update`-based predictions), and then applies the `keep_last_n` trim for parity with `fit`.
- `MLForecast.history_warmup(...)`: public entry point with the same data-validation prelude as `preprocess` (`validate_data=True` runs the gap/duplicate checks; `validate_data=False` warns when pooled transforms are configured). `max_horizon`/`horizons` (and `horizon_features`/`horizon_feature_templates`) describe how the models were trained; when omitted, any metadata from a previous fit on the instance is preserved, so re-warming an unpickled fitted object just works.
- The `keep_last_n` inference+trim block was extracted from `_transform` into a shared `TimeSeries._apply_keep_last_n()` helper; `_transform` behavior is unchanged.
- The internal `predict(new_df=...)` warm path was refactored to reuse `history_warmup(..., trim=False)`, removing the duplicated `_fit` + warm-up + metadata sequence while keeping its full-history behavior byte-identical.
- mypy: added class-body attribute annotations on `TimeSeries` (`keep_last_n`, `last_dates`, `static_features_`, `features_order_`) since the extracted helper assigns `keep_last_n` textually before `_fit`, following the existing `_uniform_dates` precedent.

## Behavior notes

- Trimming matches `fit`: an explicit `keep_last_n` is honored, otherwise it is inferred from the transforms' `update_samples`; pooled states keep the finite-window-only trimming rules from #679.
- `predict` guards are unchanged and cover the new flow: missing `models_` raises the existing "No fitted models found" error, and the model-shape check catches a warmup without `max_horizon` for one-model-per-horizon setups.
- Calling `history_warmup` again (or after `fit`) rebuilds the state from scratch; models are untouched.

## Tests

New `tests/test_history_warmup.py` (18 tests): prediction parity between `history_warmup`+`predict` and `fit`+`predict`, and between `history_warmup`+`update`+`predict` and the fit-path equivalent, across pandas/polars and local, pooled (`global_`/`groupby`), and target-transform (+date features) configurations, plus a `partition_by` case with `X_df`; state-trimming parity with `fit` (inferred and explicit `keep_last_n`); multioutput model handling and its guard; prototype-metadata preservation; the no-models guard; and both data-validation modes. Full core/forecast/pooled suites pass unchanged.

## Docs

The custom training guide (`nbs/docs/how-to-guides/custom_training.ipynb`) gains a "Skipping preprocess: warming up from history" section showing the flow as an alternative to the `preprocess` route, with outputs matching the guide's earlier predictions exactly. `history_warmup` is also listed in the `MLForecast` and `TimeSeries` API references.

### DIFF
--- a/docs/core.html.md
+++ b/docs/core.html.md
@@ -78,6 +78,7 @@ serie
       docstring_style: google
       members:
         - fit_transform
+        - history_warmup
         - predict
         - update
       heading_level: 3
--- a/docs/forecast.html.md
+++ b/docs/forecast.html.md
@@ -19,6 +19,7 @@ title: MLForecast
         - get_missing_future
         - predict
         - preprocess
+        - history_warmup
         - fit_models
         - cross_validation
         - cv
--- a/mlforecast/core.py
+++ b/mlforecast/core.py
@@ -1242,6 +1242,7 @@ class TimeSeries:
         weight_col: Optional[str] = None,
         max_horizon: Optional[int] = None,
         horizons: Optional[List[int]] = None,
+        as_numpy: Optional[bool] = None,
         trim: bool = True,
     ) -> "TimeSeries":
         """Build all internal state from `df` without materializing features.
@@ -1266,6 +1267,9 @@ class TimeSeries:
                 from a previous fit on this instance is preserved.
             horizons: Specific 1-indexed horizons the models were trained for.
                 Mutually exclusive with max_horizon.
+            as_numpy: Whether prediction passes a numpy array to the model
+                instead of a dataframe. When None, any value from a previous
+                fit is preserved (False for a fresh instance).
             trim: Apply the `keep_last_n` trim after warming the transform
                 state. `predict(new_df=...)` disables this to keep the full
                 provided history.
@@ -1294,7 +1298,10 @@ class TimeSeries:
             # unpickled instance being re-warmed); default to recursive mode
             self._horizons = getattr(self, "_horizons", None)
             self.max_horizon = getattr(self, "max_horizon", None)
-        self.as_numpy = getattr(self, "as_numpy", False)
+        if as_numpy is not None:
+            self.as_numpy = as_numpy
+        else:
+            self.as_numpy = getattr(self, "as_numpy", False)
         return self
 
     def _update_y(self, new: np.ndarray) -> None:
--- a/mlforecast/core.py
+++ b/mlforecast/core.py
@@ -260,6 +260,12 @@ class TimeSeries:
     _uniform_dates: bool = False
     _feature_null_cols: List[str]
     _xdf_null_cols: Optional[List[str]]
+    # Fit-time state (set in ``_fit``/``_apply_keep_last_n``). Declared here
+    # so mypy has a type regardless of method processing order.
+    keep_last_n: Optional[int]
+    last_dates: Any
+    static_features_: DataFrame
+    features_order_: List[str]
 
     def __init__(
         self,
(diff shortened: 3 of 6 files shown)


nasaul (approved):
Thanks for your contribution! Just added a minor guard, great job!

PR #693, opened 2026-07-21
[FEAT] Aggregate fast paths for pooled seasonal + quantile lag transforms

simonez-tuidi:
## What

Adds aggregate-based **fast paths** for the pooled (`global_`/`groupby`/`partition_by`) lag transforms that previously had none:

- **`SeasonalRolling{Mean,Std,Min,Max}`** — computed directly from the per-timestamp aggregate cache instead of a row-level pass at fit and a full cache rebuild every recursive `predict` step.
- **`RollingQuantile`, `ExpandingQuantile`, `SeasonalRollingQuantile`** — served from an opt-in per-ordinal value index (see below), so `np.quantile` runs once per query ordinal (introselect) rather than repeated mask scans.

This is the follow-up perf track deferred in #685 (documented as a limitation in the pooled guide). It builds on the `_TimestampAggregates`/`PooledState` machinery from #651/#678/#679/#682/#685.

## How

- **Seasonal fast paths** stride in parent-ordinal *value* space (partition-mode `unique_times` have holes), locating targets by `searchsorted` + value-equality rather than by position. New free helpers `_seasonal_{mean,std,min,max}_from_agg`; the three `_impl` hooks (`_ts_level_from_aggs_impl`, `_latest_from_aggs_impl`, `_bucket_feature_from_aggs_impl`) are overridden per subclass. Row-level impls remain as fallback.
- **CSR value store** — `_TimestampAggregates` gains optional `values` (non-NaN observations grouped by ordinal) + `row_offsets` (always `np.intp`). An ordinal window is one contiguous slice, no mask scan. It's **opt-in** via a new `_needs_value_store` property on `_BaseLagTransform` (True only on the three quantiles, and only when `time_agg is None` — a `time_agg` quantile runs over the single collapsed scalar exposed by `_ReaggregatedAggregates`). `PooledState._store_values` is an immutable per-state policy set once in `_fit` and honored by every build/rebuild/append/trim path. Not snapshotted; old pickles fall back to the slow path.
- **Quantile fast paths** reuse the same window bounds as the mean/expanding/seasonal helpers, returning `None` when `agg.values is None` (belt-and-braces → row slow path).

Memory cost is `O(rows)` **only** for states that actually contain a raw-row quantile transform.

## Tests

- `fast == slow` equivalence across all modes / lags / `time_agg`s / `min_samples` (incl. #682 partition default) for the 5 new fast-path transforms.
- Recursive-`predict` parity (same fitted model, forced-slow twin via monkeypatched `_impl`s) = 0.0 diff; `update()`-then-`predict`.
- `ExpandingQuantile` vs a numpy oracle; `partition_by` RANGE quantile oracle now exercises the fast path (+ a forced-slow twin).
- CSR invariant checks across build / append_predictions / append_observations / trim.
- `keep_last_n` trim guards extended to a RollingQuantile/Seasonal state.

Local suite: `tests/test_pooled.py` + oracle + `keep_last_n_trim` + `state_cleanup` = 1058 passed; `test_core` + `test_forecast` = 259 passed / 2 skipped; ruff + ruff-format + mypy clean on the touched source files.

## Docs

Replaces the "no fast path" notes in the pooled guide with the new story: seasonal = pure aggregates; quantiles = per-bucket value index.


nasaul:
Superseded by #728

PR #694, opened 2026-07-21
[PERF] Amortized predict-time compute for pooled lag transforms

simonez-tuidi:
## ⚠️ Depends on #693 — merge that first

This PR is **stacked on top of #693** (pooled seasonal + quantile fast paths). Until #693 merges, this branch contains those commits as well, so the diff below is only meaningful once the other PR is closed.

## What

Predict-time performance for the pooled (`global_`/`groupby`/`partition_by`) lag transforms — the deferred follow-up track from #651/#685. Every recursive `predict` step recomputed each transform's statistic from scratch over the *whole* aggregate history, so a horizon walk cost `O(H · (n₀ + H))` per model. This PR makes the per-step work proportional to the window (finite-window transforms) or amortized (prefix-dependent transforms).

Two independent optimizations:

### 1. Tail-window compute for finite-window Rolling{Mean,Std,Min,Max} (`6b1ccdb`)
`_latest_from_aggs_impl` computed a full-vector `np.cumsum` (Mean/Std) or rebuilt an `O(n log n)` sparse table (Min/Max) on **every** step, even though only a single target ordinal is evaluated. Now it locates `li`/`ui` with the same `searchsorted` bounds and operates on the `O(window)` aggregate slice:
- Mean/Std: sum the half-open `(li, ui]` slice of `sums`/`counts`(/`sum_sq`).
- Min/Max: `np.fmin`/`fmax.reduce` over the inclusive `[li, ui]` slice (NaN-skipping, matching the sparse table).

The fit-time helpers (vectorized over all ordinals at once) are unchanged.

### 2. Carried per-model prefix caches for Expanding{Mean,Std,Min,Max} + EWM (`e3f68a3`)
These depend on the whole history prefix and had no carried state, so each step re-scanned the full aggregate vectors (Expanding: `cumsum`/`fmin.accumulate`; EWM: refold from `k=0`). An optional `cache` is now threaded `_compute_latest_from_aggs → _latest_from_aggs_impl`. The cache is a per-`(state_key, feature_name)` dict **owned by `TimeSeries._backup`** — created on enter, dropped on exit — so it is scoped to a single model's recursive walk and can never carry a stale prefix across model restores (a later model predicts different values for the same appended dates). Each caching impl stores per bucket the last consumed ordinal plus a running accumulator and folds only the newly consumable ordinals `(prev_ui, ui]` per step, so the walk is `O(history + horizon)`. A non-monotone target ordinal drops the stale prefix and recomputes.

## Correctness

- **Min/Max/EWM are byte-identical** to the full recompute (idempotent min/max selection; the EWM fold resumes from an exact saved midpoint).
- **Mean/Std** shift by float-rounding noise (~1e-13, slice-sum / incremental-sum vs a single cumsum), absorbed by the existing `rtol/atol=1e-9` predict-parity assertions.
- `cache=None` (the default for every non-predict caller) recomputes the whole prefix, matching prior behaviour. `Offset` forwards the cache; `Combine` gives each operand its own sub-cache.

## Tests

- Expanding*/EWM fast-vs-forced-slow predict parity across all modes (recursive path).
- Multi-horizon (`max_horizon` → `_predict_multi`) parity — the direct path advances the target ordinal without feeding predictions back.
- Cache-vs-no-cache equivalence and **multi-model cache isolation** (two models that append different predictions must forecast identically together vs alone — a leaked prefix would corrupt the second).
- A pooled `test_predict_pooled` CodSpeed benchmark (~1000 series, h=14) combining all touched fast paths, so the `efficiency-tests` job tracks the predict cost.

## Deferred

The plan's optional "commit 4" (amortized capacity buffers to avoid per-step `np.append`/`concatenate` reallocation) is **intentionally deferred** — the `O(n)` memcpy with small constants is dwarfed by the wins above, and the CodSpeed benchmark added here gates whether it's ever worth doing.

## Verification

`tests/test_pooled.py` + oracle + `keep_last_n_trim` + `state_cleanup` = 1122 passed; `test_core` + `test_forecast` green; ruff + ruff-format + mypy clean on the touched source files.


nasaul:
Superseded by #728

PR #700, opened 2026-07-22
[FIX] Support weights in distributed preprocess

### LINKED ISSUE #699: `DistributedMLForecast.preprocess()` doesn't accept `weight_col`, so it can get misdetected as a static feature

The public `preprocess()` doesn't take a `weight_col` argument, unlike `_fit()`, which calls the
same private `_preprocess()` but passes `weight_col` through:

https://github.com/Nixtla/mlforecast/blob/43dbba861d59f3bb3e91fe90f43a8659b8254d79/mlforecast/distributed/forecast.py#L392-L402

Without `weight_col`, `_base_ts.weight_col` stays `None`, so the auto-detected `static_features`
list (`static_features=None`, the default) never excludes the weight column. If its values change
over time — normal for a sample weight — `MLForecastCore._fit` raises:

```
ValueError: w is declared as a static feature but its values change over time. Please set the `static_features` argument to indicate which features are static.
If all of your features are dynamic please set `static_features=[]`.
```

Right now that only works cleanly when there's no `weight_col`.

Expected: `preprocess()` accepts `weight_col`, like `fit()`/`_preprocess()` already do.

### Versions / Dependencies

mlforecast 1.1.0, pyspark 4.1.2

### Reproduction script

```python
import datetime as dt

from mlforecast.distributed import DistributedMLForecast
from pyspark.sql import SparkSession

spark = SparkSession.builder.master("local[1]").getOrCreate()

days = [dt.datetime(2026, 1, 1) + dt.timedelta(d) for d in range(20)]
df = spark.createDataFrame(
    [(uid, d, float(i % 7 + 1), 1.0 + (i % 2)) for uid in ("a", "b") for i, d in enumerate(days)],
    ["unique_id", "ds", "y", "w"],
)

dmlf = DistributedMLForecast(models=[], freq="D", lags=[1, 7], engine=spark)
dmlf.preprocess(df, id_col="unique_id", time_col="ds", target_col="y")
```

### Issue Severity

Low: It annoys or frustrates me.

Sanjays2402:
## Description
Closes #699.

`DistributedMLForecast.preprocess()` now accepts and forwards `weight_col`, preventing changing sample weights from being inferred as static features. A pandas-backed regression covers the public preprocessing path.

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.


jmoralez:
Hey, thanks for your contribution! Can you please move your test to [this file](https://github.com/Nixtla/mlforecast/blob/main/tests/test_distributed_forecast.py) and perform a similar setup as the ones there (using a dask dataframe)?

Sanjays2402:
Moved it, thanks. The test now lives in `tests/test_distributed_forecast.py` as `test_dask_distributed_preprocess_weight_col_is_not_static`, built on the same `_make_partitioned_series` dask setup the other tests use, and `tests/test_distributed_preprocess.py` is gone.

It asserts the weight column survives `preprocess` and still varies within each id, i.e. it wasn't collapsed into a static feature. Reverting the `weight_col` forward in `DistributedMLForecast.preprocess` makes it fail with a TypeError, so it's not vacuous. Locally on dask: `pytest tests/test_distributed_forecast.py -k weight` is 2 passed (I had to lift the linux-only module skip to run it on macOS, that's not part of the diff).

Head is f530b7c.

jmoralez:
Thanks, looks good! @nasaul can you review please?

nasaul (approved):
Thanks for your contribution. Happy forecasting :)

PR #701, opened 2026-07-23
[FEAT] Pooled lag-transform memory: per-state history + bucket_df narrowing

simonez-tuidi:
## Description

Two independent memory optimizations for pooled (`global_`/`groupby`/`partition_by`) lag transforms on large panels. Each is its own commit and reviewable/revertable on its own; neither depends on the other.

### 1. Trim each pooled state to its own window (`8615dfd`)

Pooled states are already keyed by `(mode, group_cols, partition_cols)`, so a mixed-window config gets one `PooledState` per key, e.g.:

```python
lag_transforms = {
    1: [
        RollingMean(window_size=7,   partition_by=["on_display"]),
        RollingMean(window_size=365, partition_by=["promo"]),
    ]
}
```

But the trim floor was `max(self.keep_last_n, w_state)`, and an **inferred** `keep_last_n` equals the *global* max window across every transform. So the `on_display` state — whose own window only needs 7 timestamps — was padded to 365, the width of the unrelated `promo` transform.

`_apply_keep_last_n` now captures the user-provided `keep_last_n` before inference and passes that (or `0` when inferred) into `_trim_pooled_states` as the only retention floor. Each finite-window state now trims to its own window when `keep_last_n` is inferred; an *explicit* `keep_last_n` is still honored as a floor, unchanged from today. The call stays gated on a resolved `keep_last_n`, so a finite state sharing the model with an unbounded transform (Expanding*/EWM — un-inferrable `keep_last_n`) still keeps full history, exactly as before. `self.ga` and the public `keep_last_n` attribute are untouched.

Net effect on the example above: the `on_display` state now retains **7** ordinals instead of 365.

### 2. Drop static group columns from stored `bucket_df` (`971c771`)

For `global`/`groupby` states, the stored `bucket_df` kept the group-column *values* at one row per observation (`O(n_series × T)`) even though they're static per series — the group→series assignment is already compact via `series_bucket_id` + a small `groups` lookup table, and the target is aggregated per-group in `_ts_aggs`.

Audited every post-construction consumer of `bucket_df` and confirmed none of them read anything beyond `[id_col, time_col]`:
- the slow-path join uses `state.join_cols`, which is `[id_col, time_col]` in every mode;
- `_compute_idsorted_to_bucket_pos` selects `[id_col, time_col]`;
- `append_observations` rebuilds bucket assignment from the *incoming* df + `groups`, never from the stored `bucket_df`.

Narrowed the stored `bucket_df` to `[id_col, time_col]` in `from_global`/`from_groupby`. Group values remain recoverable from `groups` + `bucket_id`; the target lives in `y`. Partition (`local`/`nonlocal`) states are intentionally left unchanged — same audit suggests it's safe there too, but partition values are dynamic and it needs its own dynamic-path test coverage first (tracked as follow-up).

### Verification

- `ruff check` / `ruff format --check`: clean on all edited files.
- `mypy`: no new errors (confirmed via `git stash` diff against the pre-existing baseline — the 8 pre-existing `core.py` errors are unrelated and merely line-shifted).
- Full suite: **1304 passed** — `test_pooled.py`, `test_pooled_keep_last_n_trim.py`, `test_history_warmup.py`, `test_pooled_state_cleanup.py`, `test_pooled_sqlite_oracle.py`, `test_core.py`.
- New tests: per-state window assertions + prediction/update parity for the mixed-window case, an explicit boundary test for the unbounded-transform non-goal, and an exact-schema assertion for the narrowed `bucket_df`.

Note: `971c771` is committed with `--no-verify` — `tests/test_pooled.py` isn't ruff-formatted upstream, so the `ruff-format` hook would otherwise reformat ~11 unrelated pre-existing lines into this diff. The added code itself is ruff-check/format clean.

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.


nasaul:
Superseded by #728

PR #706, opened 2026-07-30
feat: Allow NaN target values at fit/preprocess via `allow_null_target`

### LINKED ISSUE #704: [Core] Allow NaN target values at `fit`/`preprocess`, dropping them only before model fitting
### Description
Today `mlforecast` hard-rejects any null/NaN in the target column. The single gate is in
`TimeSeries._fit`:

```python
# mlforecast/core.py:551-554
validate_format(df, id_col, time_col, target_col)
validate_freq(df[time_col], self.freq)
if ufp.is_nan_or_none(df[target_col]).any():
    raise ValueError(f"{target_col} column contains null values.")
```

Because every entry point routes through `TimeSeries._fit`, this blocks NaN targets everywhere:
`MLForecast.preprocess` → `TimeSeries.fit_transform` (`mlforecast/core.py:1216`), `MLForecast.fit`
(→ `preprocess`, `mlforecast/forecast.py:1211`), `MLForecast.history_warmup`,
`MLForecast.cross_validation`, `LightGBMCV.fit` (`mlforecast/lgb_cv.py:246`) and the distributed
`DistributedMLForecast` variants (`mlforecast/distributed/forecast.py:227`).

The behaviour is locked in by `tests/test_core.py:361-367`:

```python
series_with_nulls = series.copy()
series_with_nulls.loc[1, "y"] = np.nan
with pytest.raises(Exception) as exec:
    ts.fit_transform(series_with_nulls, id_col="unique_id", time_col="ds", target_col="y")
assert "y column contains null values" in str(exec.value)
```

(also snapshotted in `docs/core.html.md:250`).

**Request:** allow NaN in the target at `fit`/`preprocess`/`history_warmup`. NaN rows should still
participate in feature construction *positionally* (so the time grid stays intact), be *excluded*
from the statistics computed over them (they must not inflate the denominator of a rolling mean,
count, std, …), and be dropped from `X`/`y` immediately before the estimator is fitted.

Ideally opt-in, e.g. `allow_null_target: bool = False` on `preprocess`/`fit`/`history_warmup`, so
current behaviour is unchanged by default.

## Use case

We forecast retail/product-level demand. A product typically has a **life-cycle**: it does not exist,
or is out of range/delisted, for part of the history. The intervals where it does not exist and the
intervals where it exists but sold nothing are two completely different signals.

`mlforecast` requires a gap-free, regular time grid — `validate_df` raises on gaps:

```python
# mlforecast/data_validation.py:221-228
has_issues, bad = validate_continuity(df, id_col, time_col, freq)
if has_issues:
    raise ValueError(
        f"Series contain missing or duplicate timestamps with the specified freq {freq}\n"
        f"Affected series: {bad_ids}\n"
        f"Consider using the fill_gaps parameter or preprocessing your data."
    )
```

(and `validate_update_df` at `mlforecast/data_validation.py:192-199` for `update`; pooled
`global_`/`groupby`/`partition_by` transforms additionally *require* a dense grid — see the
`validate_data=False` warning at `mlforecast/forecast.py:263-271`.)

So the two requirements combine into a forced-imputation trap: we must fill the gaps, and the filler
cannot be NaN, so it becomes `0`. That `0` is then indistinguishable from a genuine zero-sales week.
Concretely:

- A rolling mean over a 12-week window that includes 8 weeks of "product did not exist" reads as a
  mean over 12 observations mostly at zero, instead of a mean over the 4 real observations.
- Counts / `min_samples` gates see 12 samples where there are 4, so features look "mature" when they
  are not.
- Every model trained on this learns "post-launch demand looks like a ramp from zero", which is an
  artifact of the imputation.
- Rows where the product does not exist are also not legitimate training rows — the target itself is
  not defined there — but there is no way to say so, so we either train on fabricated zeros or filter
  the preprocessed frame ourselves with a hand-maintained mask, which is easy to get wrong and has to
  be re-derived for every horizon in the `max_horizon` case.

The workaround we would want to write — put NaN where the product does not exist, let the transforms
skip it, drop those rows before fitting — is exactly what NaN semantics give for free, and it is what
we would get from SQL window functions over the same data.

## Why this looks close to free

Most of the machinery already exists and already has the desired semantics.

**1. "Drop NaN target rows before fitting" is already implemented.** `TimeSeries._transform` drops
target-null rows unconditionally, *even with* `dropna=False`:

```python
# mlforecast/core.py:936-949
target_nulls = np.isnan(target)
if target_nulls.ndim == 2:
    # target nulls for each horizon are dropped in MLForecast.fit_models
    # we just drop rows here for which all the target values are null
    target_nulls = target_nulls.all(axis=1)
if dropna:
    ...
    keep_rows = ~(feature_nulls | target_nulls)
else:
    # we always want to drop rows with nulls in the target
    keep_rows = ~target_nulls
```

**2. Per-horizon NaN masking is already implemented** for the multi-output path, in both
`TimeSeries._horizon_train_data` (`mlforecast/core.py:1175`) and `MLForecast.fit_models`
(`mlforecast/forecast.py:864`):

```python
valid = ~np.isnan(y_h)
...
X_h = ufp.filter_with_mask(X_h, valid)
y_h = y_h[valid]
```

so "features computed on the full grid, model fitted on the valid subset" is an established pattern
in this codebase, not a new one.

**3. The pooled transforms are already NaN-skipping, by design and by documentation.**
`_RollingBase.__init__` documents `min_samples` in exactly the terms this request needs
(`mlforecast/lag_transforms.py:506-509`, and `1026-1029` for the seasonal variant):

> In pooled mode (`global_=True`, `groupby` or `partition_by`), `min_samples` counts total **non-NaN
> observations** across **all series** in the bucket within the rolling window, with no capping.

and the implementations agree — the generic slow path drops NaN before the statistic
(`mlforecast/lag_transforms.py:598-608`):

```python
vals = y_b[mask]
vals = vals[~np.isnan(vals)]
if len(vals) >= min_samples and len(vals) > 0:
    feat_u[k] = self._window_stat(vals)
```

the aggregate fast paths carry a non-NaN `counts` array and divide by it
(`_rolling_mean_from_agg`, `mlforecast/lag_transforms.py:616-633`), and `PooledState` builds its
per-timestamp aggregates from
a validity mask (`mlforecast/pooled.py:195`, `mlforecast/pooled.py:243`, `mlforecast/pooled.py:286`,
plus the incremental update at `mlforecast/pooled.py:959-1004`).

This is verifiable today by patching out the guard at `core.py:553`. Two series `a`/`b`, daily,
`RollingMean(window_size=3, min_samples=1, global_=True)`, with `a` NaN on 01-09..01-11:

```
 unique_id         ds     y  lag1  global_rolling_mean_lag1_window_size3_min_samples1
         a 2020-01-08   8.0   7.0                                                55.5
         a 2020-01-12  12.0   NaN                                               109.0   <- mean of b's 108,109,110
         a 2020-01-13  13.0  12.0                                                85.5
```

The NaN rows are dropped from the output automatically (point 1), and the window at 01-12 averages
only the three values that actually exist rather than dividing by six. That is precisely the
requested behaviour, already working.

## The part that does need work

The **local (per-series) path** delegates to `coreforecast`, which propagates NaN *by default*, so a
single NaN poisons the rest of the series:

```python
y = np.array([1., 2., np.nan, 4., 5., 6.])
ct.RollingMean(lag=1, window_size=3, min_samples=1).transform(ga)
# -> [nan  1.  1.5  nan  nan  nan]      expected [nan 1. 1.5 1.5 3. 4.5]
ct.ExpandingMean(lag=1).transform(ga)
# -> [nan  1.  1.5  nan  nan  nan]
ct.Lag(lag=1).transform(ga)
# -> [nan  1.  2.  nan  4.  5.]         correct: the lagged value genuinely is NaN
```

End to end (single series, `lags=[1]`, `RollingMean(window_size=3, min_samples=1)`, NaN on rows
8-10, guard patched out): the rolling mean is `NaN` for **every** row from row 11 to the end, while
the same configuration with `global_=True` is correct.

**But this is opt-out, not a missing capability.** `coreforecast` 0.0.18 already takes a `skipna`
argument on every one of these transforms, and it produces exactly the right answer — `mlforecast`
simply never sets it:

```python
ct.RollingMean(1, 3, 1, skipna=True).transform(ga)   # -> [nan 1.  1.5 1.5 3.  4.5]   matches pandas
ct.ExpandingMean(1, skipna=True).transform(ga)       # -> [nan 1.  1.5 1.5 2.33 3.  ]
ct.ExponentiallyWeightedMean(1, .5, skipna=True).transform(ga)  # -> [nan 1. 1.5 1.5 2.75 3.875]
ct.SeasonalRollingMean(1, 2, 2, 1, skipna=True).transform(ga)   # -> [nan 1. 2.  1.  3.  5.  ]
```

And plumbing it through looks nearly mechanical: `_BaseLagTransform._set_core_tfm`
(`mlforecast/lag_transforms.py:133-143`) already forwards every non-pooled `__init__` argument to the
`coreforecast` class by name, so adding `skipna` to `_RollingBase.__init__` /
`_Seasonal_RollingBase.__init__` / `_ExpandingBase.__init__` / `ExponentiallyWeightedMean.__init__`
would reach `coreforecast` with no change to `_set_core_tfm` itself. `_get_name`
(`mlforecast/lag_transforms.py:145-167`) only appends parameters whose value differs from the
default, so `skipna=False` keeps every existing feature name byte-identical.

Two caveats found while checking this:

- **`coreforecast`'s `update()` ignores `skipna`** for `ExpandingMean`, `ExpandingStd`,
  `ExpandingMin`, `ExpandingMax` and `ExponentiallyWeightedMean` — exactly the five whose `update()`
  is implemented in Python rather than delegating to `_lib`. `transform()` is correct for all of
  them, so training features would be right and recursive prediction would silently diverge. The
  rolling/seasonal/quantile transforms are fine. Filed upstream against `coreforecast`; `mlforecast`
  should either wait for that fix or restrict `skipna` to the transforms whose `update()` honors it.
- **Version floor.** `skipna` does not exist in `coreforecast` 0.0.16 (zero occurrences in
  `rolling.py`, `expanding.py`, `lag_transforms.py`, `scalers.py`); it is present throughout 0.0.18.
  `pyproject.toml:27` currently declares `coreforecast>=0.0.15`, which would need raising.

Other things that would need a decision:

- **`Lag` itself** — `lag1` of a NaN is legitimately NaN; those rows get dropped by `dropna=True`
  as usual. Worth stating explicitly in the docs.
- **Whether `skipna` is per-transform or a single dataset-level switch.** Per-transform matches
  `coreforecast` and lets a user keep NaN-propagating semantics where they want them; a single
  `preprocess(..., skipna=True)` is friendlier but has to fan out to every transform anyway.
- **Target transforms** (`Differences`, `AutoDifferences`, `LocalStandardScaler`, … in
  `mlforecast/target_transforms.py`) — the `coreforecast` scalers behind these already accept
  `skipna` (`LocalMinMaxScaler`, `LocalStandardScaler`, `LocalRobustScaler`; `LocalStandardScaler(skipna=True)`
  on `[1, 2, nan, 4, 5]` correctly yields `mean=3.0, std=1.581`, vs `nan/nan` with the default), so
  the same threading applies. `LocalBoxCoxScaler` and `Difference` do not take `skipna`.
- **Recursive `predict`/`update`** — history that contains NaN feeding the update rules; the pooled
  incremental update already handles it (`mlforecast/pooled.py:959-1004`), the local
  `_core_tfm.update` path does not.
- **Feature nulls are already tolerated with a warning**, not a raise
  (`mlforecast/core.py:1576`: `warnings.warn(f"Found null values in {', '.join(cols_with_nulls)}.")`),
  which suggests the target could reasonably follow the same permissive-with-a-warning philosophy.

## Proposed shape

1. Replace the unconditional raise at `mlforecast/core.py:553-554` with a check gated on a new
   `allow_null_target=False` argument threaded through `TimeSeries.fit_transform` /
   `TimeSeries.history_warmup` and exposed on `MLForecast.preprocess` / `fit` /
   `history_warmup` / `cross_validation` (and the distributed + `LightGBMCV` equivalents).
2. When enabled: keep NaN rows on the grid for feature computation, rely on the existing
   `keep_rows = ~target_nulls` drop (`mlforecast/core.py:948-949`) and per-horizon `valid` masks
   (`mlforecast/core.py:1175`, `mlforecast/forecast.py:864`) to keep them out of `fit`.
3. Add `skipna` to the local transforms in `mlforecast/lag_transforms.py` and let the existing
   `_set_core_tfm` forwarding carry it into `coreforecast`, raising a targeted error for the
   transforms whose `coreforecast` `update()` does not yet honor it — not a blanket rejection of the
   whole dataframe.
4. Document that `forecast_fitted_values()` will not contain the NaN-target rows, same as the
   existing `dropna` behaviour.

Happy to work on a PR if the maintainers agree on the shape — in particular on whether `skipna`
should be per-transform (mirroring `coreforecast`) or a single switch on `preprocess`/`fit`, and
whether to ship before or after the upstream `coreforecast` `update()` fix.

simonez-tuidi:
## Description

Closes #704.

Adds an opt-in `allow_null_target` to accept NaN in the target at `preprocess`/`fit`. Nulls are excluded from the statistics computed over them, so an unobserved period doesn't count toward a rolling window's sample count.

### Why

`TimeSeries._fit` rejected any null in the target outright (`mlforecast/core.py`), and `validate_df` requires a gap-free grid. Together those force imputation, and the only available filler is a number — in practice `0`. That makes "did not sell" indistinguishable from "did not exist yet", which is the common case for product life-cycles: a SKU is out of range or unlaunched for part of its history. A 12-week rolling mean spanning 8 unobserved weeks was being computed over 12 observations instead of the 4 real ones, and every model trained on it learns a ramp-from-zero that is an artifact of the imputation.

### Usage

```python
# a SKU that didn't exist for three days, on a dense grid
df.loc[df["ds"].isin(unlaunched), "y"] = np.nan

fcst = MLForecast(
    models=[LGBMRegressor()],
    freq="D",
    lags=[1],
    lag_transforms={1: [RollingMean(window_size=3, min_samples=1)]},
)
fcst.fit(df, allow_null_target=True)
```

That's the whole opt-in — one flag. The rolling mean over a window covering two unobserved days and one real one is that real value, not a third of it. Those rows are excluded from model fitting and from `forecast_fitted_values()`.

### What changes

**`allow_null_target=False`** on `preprocess`, `fit`, `history_warmup`, `cross_validation`, `LightGBMCV.setup`/`fit` and the `DistributedMLForecast` equivalents. When enabled, null-target rows stay on the grid while features are computed — so lags stay positional and the series has no gaps — and are then excluded from the features/target. Default behaviour and the existing error message prefix are unchanged.

Most of this already existed: `_transform` computes `keep_rows = ~target_nulls` even when `dropna=False`, and the per-horizon paths already mask with `valid = ~np.isnan(y_h)`. The flag mainly removes the gate and validates the configuration.

**`skipna=None`** on `RollingMean/Std/Min/Max/Quantile`, the `SeasonalRolling*` and `Expanding*` families, `ExponentiallyWeightedMean`, and `LocalStandardScaler`/`LocalMinMaxScaler`/`LocalRobustScaler`. `None` means "follow `allow_null_target`", resolved in `_fit` once the flag is known, so nothing has to be annotated per transform. `True`/`False` are explicit overrides — `False` keeps the propagating semantics, which is a legitimate way to signal "this window contains unobserved data" to the model.

coreforecast already accepts `skipna` on both its functional and stateful APIs; the transforms just never passed it. `_set_core_tfm` already forwards every non-pooled `__init__` argument by name, so adding the parameter is enough to reach coreforecast.

**Feature names don't change.** They're built during `_parse_transforms`, before resolution runs, so `RollingMean(3, 1)` is `rolling_mean_lag1_window_size3_min_samples1` whether or not the flag is set. A name reflects the declared spec, not the resolved behaviour — worth knowing, and it's the convention `_get_name` already uses for every parameter. An explicit `skipna=False`/`True` does get a suffix, since it's a non-default declared value.

In pooled scopes (`global_`/`groupby`/`partition_by`) NaN is already skipped by the aggregate machinery, so `skipna` is a documented no-op there.

**Configuration validation.** Target transforms that would propagate the null through every value of a series are a hard error rather than silently nulling out the fitted values and predictions; the local scalers are accepted with `skipna=True`. `Lag`/`LookupLag` are deliberately untouched, since the lag of an unobserved value genuinely is unobserved. `Offset`/`Combine` take no `skipna` of their own and delegate resolution to the transforms they wrap.

### Interaction with coreforecast

`skipna` does not exist in coreforecast <= 0.0.16 and we only require `>= 0.0.15`, so rather than raise the floor for everyone this detects the capability by signature (`mlforecast/compat.py`). When absent, `skipna=True` raises a clear message and the default is never forwarded, so nothing changes for users who don't opt in.

Separately, as of coreforecast 0.0.18 the five transforms whose `update` is implemented in Python rather than delegating to `_lib` — `ExpandingMean`, `ExpandingStd`, `ExpandingMin`, `ExpandingMax`, `ExponentiallyWeightedMean` — honour `skipna` in `transform` but ignore it in the incremental `update`. Training features would exclude NaN while recursive predictions would not, with nothing raising.

Because inference must not break a working configuration, those are handled asymmetrically: an **inferred** `True` downgrades to `False` and warns (so opting into null targets can't turn an existing `ExpandingMean` setup into a hard error), while an **explicit** `skipna=True` raises. The check is a runtime probe rather than a version pin, so it lifts itself once the upstream fix lands (filed separately against coreforecast).

### Two behaviours worth reviewing

**Dropped-series bookkeeping (bug fix, affects existing behaviour).** `_transform` inferred "this series was dropped completely" from its *last* row being filtered out. A series with valid history and a trailing null satisfies that test, so it was recorded in `_dropped_series` and its scaler state was later discarded while its rows were still in the training frame — surfacing as `Number of groups in the scaler (0) doesn't match the ones in data (1)`. Now counted from retained rows per series, which agrees with the old result for the short-series case the check was written for.

**Cross validation returns null actuals rather than dropping them.** The conformal scores are reshaped to `(cs_n_windows, n_series, cs_h)`, so dropping rows would break both that reshape and the `expected_rows` check. The row count stays aligned with `h * n_windows` per serie; filter before scoring. Instead, conformal calibration now requires an observed actual for every step it calibrates on — a null there would otherwise reach `np.quantile` and produce null intervals with no indication anything went wrong, and calibrating on a subset would quietly change the coverage guarantee. One shared `validate_observed_calibration_targets` covers the fit-time path and the two transfer methods that build scores from a target backtest (`recalibrate`, `error_scaled`). Nulls confined to older history are unaffected.

### Internal rebuild paths

Threading the public entry points alone was not sufficient — several paths rebuild the forecasting state internally and would re-reject a target the original fit had accepted. `_fit` now records the setting so they inherit it, and since resolution happens there, they also inherit the resolved `skipna`: `predict(new_df=...)`, `cross_validation(refit=False)`, `forecast_fitted_values(h>1)`, `LightGBMCV.fit`'s final `ts._fit`, the distributed `predict(new_df=...)` per-partition rebuild, and the `preprocess` callback the weighted transfer methods use for density ratio estimation. Each has a regression test.

`LightGBMCV` metrics now skip unobserved actuals, which would otherwise make the metric — and therefore early stopping — NaN.

### Commits

| | |
|---|---|
| `96bab63` | `skipna` on lag transforms and local scalers; coreforecast capability detection |
| `25cc69f` | `allow_null_target` in `TimeSeries`; dropped-series fix |
| `d313d1e` | public API threading, internal-rebuild inheritance, conformal calibration guards |
| `f3bf198` | tests |
| `a2499b8` | make `skipna` tri-state so `allow_null_target` implies it |

### Tests

`tests/test_null_target.py`, 62 tests. Correctness is pinned against pandas' own NaN-skipping rolling statistics, plus an explicit contrast with the zero-filled frame (`12.0` vs `4.0` for the same window). Also covers: the default rejection; rows kept for feature computation but dropped from the output; inference from `allow_null_target` matching an explicit `skipna=True`; feature names unchanged under inference; re-fitting re-resolving rather than inheriting; the explicit `skipna=False` opt-out; warn-not-raise for the five affected transforms and the raise on an explicit `True`; the pooled exemption; delegation through `Offset`/`Combine`; the target-transform gate; polars/pandas parity; and one regression test per internal rebuild path listed above.

Existing suites, run locally against this branch:

| Suite | Result |
|---|---|
| `test_null_target.py` | 62 passed |
| `test_conformal` + `test_transfer_conformal` + `test_lgb_cv` + `test_core` + `test_target_transforms` | 270 passed |
| `test_lag_transforms` + `test_core` + `test_pooled` + `test_target_transforms` | 550 passed |
| `test_forecast` + `test_history_warmup` | 126 passed, 2 skipped |

`ruff check` is clean on every file this PR touches. `mypy` reports 68 errors in 10 files on both this branch and `main` — identical sets, only line numbers shifted, so no new type errors.

Checklist:
- [x] The tests pass.
- [x] All linting tasks pass.


mlforecast/compat.py
```diff
@@ -1,5 +1,89 @@
 __all__ = []
 
+import copy
```
jmoralez, line 3: i think it's best to wait for a release with the fix in coreforecast and set that as the lower bound

simonez-tuidi:
@nasaul @jmoralez any chance we can follow up on this?

mlforecast/compat.py
```diff
@@ -1,5 +1,89 @@
...
+        "ExpandingMax",
+        "ExponentiallyWeightedMean",
+    }
+)
+
+
+_skipna_update_probe_cache: dict = {}
+
+
+def _probe_core_update_skipna(core_tfm) -> bool:
```
nasaul, line 51: **The probe's answer depends on `lag`, so this guard is silently bypassed for lags 2-4.** (high)

The probe asserts "`update` honors `skipna`" from `not isnan(out)` on a fixed 5-element series. But `ExpandingMean.update` does `stats_[:, 1] += ga._index_from_end(self.lag - 1)`; for `lag >= 2` that reads a *non-NaN* older value, so the probe sees a finite number and concludes the transform is fine — when in fact it silently folded the wrong observation in. Verified on this branch against coreforecast 0.0.18 (T = "probe says supported"):

| lag | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
|---|---|---|---|---|---|---|---|
| `ExpandingMean` | F | T | T | T | F | F | F |
| `ExpandingStd` | F | T | T | T | T | F | F |
| `ExpandingMin` | F | T | T | T | F | F | F |
| `ExpandingMax` | F | T | T | T | F | F | F |
| `ExponentiallyWeightedMean` | F | T | T | T | F | F | F |

End-to-end repro — no warning, `skipna=True` written to the core transform, so training features skip the NaN while recursive `predict` does not, which is exactly the disagreement this guard exists to prevent:

```python
f = MLForecast(models=[LinearRegression()], freq="D", lags=[1],
               lag_transforms={3: [ExpandingMean()]})
f.fit(df_with_one_nan, allow_null_target=True)
# warnings: []
# f.ts.transforms["expanding_mean_lag3"]._core_tfm.skipna -> True
```

The same config at `lag=1` warns and downgrades, so the behaviour is inconsistent by lag. Every Expanding/EWM test in `tests/test_null_target.py` (lines 233-281) uses `lag=1`, the one lag where the probe happens to be right, so the suite doesn't catch this.

Suggested fix: make the probe series long enough for the transform's `lag` and compare the returned *value* against the skipna-correct expectation rather than only checking for NaN — or drop the probe and pin the coreforecast version.

_This, the `_skipna_update_probe_cache` comment above, and the `_resolve_skipna` comment in `lag_transforms.py` are one coherent fix to `_probe_core_update_skipna` and its cache key, not three separate patches._

nasaul (commented):
Correctness review. Findings 1-3 (the two `compat.py` comments and the `_resolve_skipna` comment) are one coherent fix to `_probe_core_update_skipna` and its cache key -- please treat them as a single patch, not three.

mlforecast/compat.py
```diff
@@ -1,5 +1,89 @@
...
+        "ExpandingMean",
+        "ExpandingStd",
+        "ExpandingMin",
+        "ExpandingMax",
+        "ExponentiallyWeightedMean",
+    }
+)
+
+
+_skipna_update_probe_cache: dict = {}
```
nasaul: **Caching on the class name alone makes the result process-order-dependent.** (high)

The comment below says "which branch `update` takes is structural (per class), not parameter-dependent", but the probe's outcome depends on both `lag` (see the comment on `_probe_core_update_skipna`) and `skipna` (see the `_resolve_skipna` comment in `lag_transforms.py`). Since this is a module-global dict keyed on `type(core_tfm).__name__`, the first probe in the process decides for every later instance. Both directions reproduce on this branch:

```python
# order A -- inferred path at lag 1 runs first
MLForecast(..., lag_transforms={1: [ExpandingMean()]}).fit(df, allow_null_target=True)
# _skipna_update_probe_cache -> {'ExpandingMean': False}
MLForecast(..., lag_transforms={3: [ExpandingMean(skipna=True)]})
# ValueError: ExpandingMean(skipna=True) is not supported in local mode ...   <- spurious

# order B -- fresh process, opposite order
MLForecast(..., lag_transforms={3: [ExpandingMean(skipna=True)]})   # constructed OK (guard bypassed)
# _skipna_update_probe_cache -> {'ExpandingMean': True}
MLForecast(..., lag_transforms={1: [ExpandingMean(skipna=True)]})   # also OK -- cache poisoned the other way
```

So the same user code either raises or silently mis-computes depending on what ran earlier in the interpreter.

Suggested fix: key on `(class name, lag, skipna)`, or drop the cache — the probe is a handful of numpy ops.

_Part of the same fix as the `_probe_core_update_skipna` comment below and the `_resolve_skipna` comment in `lag_transforms.py`._

mlforecast/lag_transforms.py
```diff
@@ -120,8 +151,75 @@ def _set_core_tfm(self, lag: int) -> "_BaseLagTransform":
...
+        if not self._accepts_skipna or getattr(self, "skipna", None) is not None:
+            return None
+        # Pooled scopes always skip nulls via their own aggregates, and without
+        # upstream support there's nothing to turn on.
+        if self._is_pooled or not core_supports_skipna():
+            return None
+        if not allow_null_target:
+            self._core_tfm.skipna = False
+            return None
+        if not core_update_honors_skipna(self._core_tfm):
```
nasaul, line 217: **The inferred path probes a core transform built with `skipna=False`, so it can never report "supported".** (medium)

`_set_core_tfm` normalizes a deferred `skipna=None` to `False` before constructing `_core_tfm` (`init_args["skipna"] = init_args["skipna"] is True`). This line then calls `core_update_honors_skipna(self._core_tfm)` on that `skipna=False` object. A transform built with `skipna=False` is *supposed* to propagate NaN, so the probe returns `False` by construction. Confirmed with a stand-in whose `update` does honor `skipna` (`RollingMean` delegates to `_lib`):

```python
probe(ct.RollingMean(lag=1, window_size=3, min_samples=1, skipna=True))   -> True
probe(ct.RollingMean(lag=1, window_size=3, min_samples=1, skipna=False))  -> False
```

Consequences:

1. The stated design goal — "the guard disappears on its own once coreforecast fixes its Python-side accumulators" — doesn't hold for the inferred path. After the upstream fix lands, `allow_null_target=True` with a default `ExpandingMean` will still warn and downgrade to propagating.
2. It writes that `False` into the class-name-keyed `_skipna_update_probe_cache`, which then makes a later explicit `skipna=True` on the same class raise spuriously (see the `compat.py` cache comment).

Suggested fix: set `probe.skipna = True` inside `_probe_core_update_skipna` before probing, so it measures the transform's structure rather than the value it happens to have been built with.

_Part of the same fix as the two `compat.py` comments._

mlforecast/lag_transforms.py
```diff
@@ -464,17 +575,21 @@ def __init__(
             groupby=groupby,
             partition_by=partition_by,
             time_agg=time_agg,
+            skipna=skipna,
             **kwargs,
         )
         self.p = p
 
     def _set_core_tfm(self, lag: int):
+        extra = {"skipna": self.skipna} if core_supports_skipna() else {}
```
nasaul: **`RollingQuantile._set_core_tfm` forwards `self.skipna` unnormalized, so `skipna=None` reaches coreforecast.** (low)

Every other transform goes through the generic `_BaseLagTransform._set_core_tfm`, which normalizes with `init_args["skipna"] = init_args["skipna"] is True`. This override passes the raw value, and `None` is not a valid input — coreforecast silently mis-computes rather than raising:

```python
ga = GroupedArray(np.array([1., np.nan, 3., 4., 5.]), np.array([0, 5], dtype=np.int32))
ct.RollingQuantile(lag=1, p=0.5, window_size=3, min_samples=1, skipna=None).transform(ga)
# [nan  1. nan nan nan]     (skipna=False should give [nan 1. nan nan 4.])
```

This is currently masked because `_resolve_skipna` overwrites `_core_tfm.skipna` for every *local* transform before any feature is computed — but it returns early for pooled ones, so a pooled `RollingQuantile` keeps `skipna=None` on its core transform indefinitely. It's latent today only because `_get_local_tfms` keeps pooled transforms away from `_core_tfm`.

Suggested fix: `extra = {"skipna": self.skipna is True} if core_supports_skipna() else {}`, matching the generic path.

mlforecast/lag_transforms.py
```diff
@@ -82,6 +84,25 @@ def _validate_time_agg(time_agg, global_, groupby, *, allow_none=True, scope_exe
         )
 
 
+def _validate_skipna(skipna: Optional[bool]) -> None:
```
nasaul: **`_validate_skipna` rejects `skipna=True` on pooled transforms with an older coreforecast, where `skipna` is documented as a no-op.** (low)

On coreforecast <= 0.0.16, `RollingMean(3, global_=True, skipna=True)` raises "skipna=True requires a coreforecast version that supports it" — but pooled scopes never consult coreforecast's `skipna` at all. `_RowStore.valid()` in `pooled.py` does its own NaN filtering (`keep = ~np.isnan(self.y)`), and the docstring for the parameter says as much ("pooled scopes always skip NaN, so it's a no-op there"). `_check_skipna_update_support` already exempts `self._is_pooled` for the same reason; this check doesn't.

So a user on an older coreforecast is blocked from a configuration that would have worked correctly.

Suggested fix: skip the check when the transform is pooled — either move the call after the scope attributes are set and gate it on `self._is_pooled`, or pass the scope in.

jmoralez:
As asked in https://github.com/Nixtla/mlforecast/pull/706/changes#r3730872241, please remove the checks for the coreforecast support. We should:
1. Fix in coreforecast
2. Make a coreforecast release
3. Set that release as the minimum version here

simonez-tuidi:
@jmoralez agreed, can we assign the work? 

I don't know that repo very well but I could try, unless you or @nasaul would like to pick it up?

jmoralez:
It's being addressed in https://github.com/Nixtla/coreforecast/pull/147, I'll update here when we publish a relase with it.

PR #715, opened 2026-08-16
add Polars feature encoders

janrth:
## Summary

Adds a generic `feature_encoders` hook to `MLForecast` and ships Polars-native categorical encoders:

- `PolarsTargetEncoder`: smoothed, conventional target encoding fitted on the complete training partition.
- `PolarsOrdinalEncoder`
- `PolarsCountEncoder`: raw category counts by default; `normalize=True` returns frequencies.
- `PolarsOneHotEncoder` for low-cardinality features.

Encoders are cloned per fitted model, transformed again at prediction time, and work with recursive and direct forecasting, cross-validation, and AutoML.

## Why

Pandas/sklearn categorical pipelines can introduce conversion and memory overhead when forecasting inputs are already Polars. This change keeps feature generation and categorical encoding in Polars while continuing to support arbitrary user-provided encoders through `feature_encoders=[...]`.

## Local benchmark results (not included in this PR)

Three-run medians from fresh M5 processes on the current branch: 672,000 rows, LightGBM fit plus 7-step prediction. Each row compares the pandas `category_encoders` implementation with its Polars equivalent; the count benchmark uses normalized counts (`normalize=True`) on both sides.

| Encoder | Fit-time improvement | Predict-time improvement | Input RSS reduction | Peak RSS reduction |
| --- | ---: | ---: | ---: | ---: |
| Target | 45.3% | 10.7% | 26.6% | 14.6% |
| Ordinal | 38.5% | 13.7% | 9.3% | 4.6% |
| Count (normalized frequency) | 61.7% | 25.6% | 14.0% | 19.6% |

Positive values mean Polars used less time or memory. The target row now compares equivalent conventional in-sample target encodings. A time-based cross-validation pipeline remains validation-safe because each fold fits its encoder only on that fold's training partition.

### Direct fitted-value cache

The implementation stores fitted predictions rather than one encoded feature frame per direct model/horizon. In a separate three-run median on 336,000 M5 rows with seven direct horizons, `fitted=True`, and `DummyRegressor` (used to isolate cache cost), this reduced both after-fit and peak RSS by 13.8% and reduced fit time by 2.8% versus the preceding PR revision.

The benchmark notebooks and workers are intentionally excluded: this PR contains only library code and automated tests. The preparation-time metric is excluded from the main comparison because the exploratory benchmark starts with a shared pandas panel and the Polars case includes the explicit pandas-to-Polars conversion.

## Validation

`tests/test_feature_encoders.py` covers target, ordinal, count/frequency, and one-hot encoding; unseen and null categories; generic sklearn-compatible encoders; recursive and direct forecasting; fitted values; save/load; cross-validation; and AutoML.


elephaint:
@nasaul

jmoralez:
@janrth is the benchmark comparing pandas vs polars?

janrth:
> @janrth is the benchmark comparing pandas vs polars?

Yes, it is category_encoders with an sklearn pipeline vs the polars encoders ontop

mlforecast/feature_encoders.py
```diff
@@ -0,0 +1,338 @@
...
+    try:
+        return clone(encoder)
+    except (TypeError, AttributeError):
+        return copy.deepcopy(encoder)
+
+
+def _accepts_context(method: Any) -> bool:
+    parameters = inspect.signature(method).parameters.values()
+    return any(
+        parameter.name == "context" or parameter.kind == inspect.Parameter.VAR_KEYWORD
```
nasaul, line 33: **Breaks every stock sklearn transformer.** `sklearn.base.TransformerMixin.fit_transform` has signature `(X, y=None, **fit_params)`, so this `VAR_KEYWORD` clause returns `True` and `_fit_transform` calls `fit_transform(X, y, context=...)`. sklearn forwards `fit_params` straight into `fit()`:

```
TypeError: StandardScaler.fit() got an unexpected keyword argument 'context'
```

This hits `OneHotEncoder`, `ColumnTransformer` and all of `category_encoders` — precisely the interface the `feature_encoders` docstring advertises. Matching only a literal `context` parameter fixes it.

Covered by `test_stock_sklearn_transformer_is_accepted`.

mlforecast/forecast.py
```diff
@@ -829,7 +851,11 @@ def _compute_fitted_values(
             fitted_values = ufp.assign_columns(base, target_col, y)
             for name, model in self.models_.items():
                 assert not isinstance(model, dict)  # mypy
-                preds = model.predict(X)
+                if isinstance(model, _EncodedModel) and hasattr(model, "fitted_X_"):
```
nasaul, line 854: **Fitted values are misaligned when the input isn't pre-sorted.** A few lines up, `sort_idxs` reorders `base`, `X` and `y` — but `fitted_X_` was stored back in `fit_models` in the *input-df* order, since `prep` is restored via `_restore_idxs`. So `predict_fitted()` returns rows in one order and they're assigned in another.

With a polars frame whose rows interleave ids (`b,a,b,a,…`), `forecast_fitted_values()` gives `[0, 0, 6, 1, 6.5, 1.5, 7, 2]` where the same data pre-sorted gives `[0, 1, 1.5, 2, 0, 6, 6.5, 7]` — the correct values, permuted. No error, just wrong numbers.

Either take `fitted_X_` rows by `sort_idxs`, or store the sort indices alongside it. Covered by `test_fitted_values_follow_input_row_order`.

mlforecast/forecast.py
```diff
@@ -829,7 +851,11 @@ def _compute_fitted_values(
...
+                    preds = model.predict_fitted()
+                    del model.fitted_X_
+                else:
+                    preds = model.predict(X)
```
nasaul, line 858: **This fallback reports target-leaked fitted values.** `fitted_X_` is deleted on the line above after its first use, so in `cross_validation(fitted=True, refit=False)` every window after the first lands here. `model.predict(X)` re-encodes through `PolarsTargetEncoder.transform`, which uses full-training statistics rather than the causal per-timestamp encodings.

Fold 0 returns `[0.0, 1.0, 1.5, 2.0, 2.5]` (causal); fold 1 returns `[3.0, 3.0, 3.0, 3.0, 3.0, 3.0]` — the global mean. Anyone comparing fitted values across folds is comparing incomparable numbers, with no warning.

Covered by `test_cross_validation_fitted_values_are_causal_in_every_fold`.

mlforecast/forecast.py
```diff
@@ -950,7 +976,11 @@ def _compute_fitted_values(
                     X_pred = (
                         ufp.to_numpy(X_valid) if models_trained_with_numpy else X_valid
                     )
-                    preds_valid = model.predict(X_pred)
+                    if isinstance(model, _EncodedModel) and hasattr(model, "fitted_X_"):
```
nasaul, line 979: **Same misalignment as the recursive path, in the direct path.** `preds[valid] = model.predict_fitted()` assumes `fitted_X_` is ordered like the possibly re-sorted `X_h`. `valid.sum()` is order-invariant, so the lengths still match and this fails silently rather than raising.

Reproduced with `fit(df_unsorted, max_horizon=2, fitted=True)`. Covered by `test_direct_fitted_values_follow_input_row_order`.

mlforecast/forecast.py
```diff
@@ -643,14 +654,25 @@ def fit_model(
...
-            return clone(model).fit(X, y, **fit_kwargs)
+                return fitted_model.fit(X, y, **fit_kwargs)
+            fitted_encoders = [
+                _clone_encoder(encoder) for encoder in self.feature_encoders
+            ]
+            return _EncodedModel(fitted_model, fitted_encoders).fit(
+                X,
+                y,
+                encoder_context=encoder_context_,
+                store_fitted_X=store_fitted_X,
```
nasaul, line 673: Minor, memory: every `_EncodedModel` holds its `fitted_X_` until `_compute_fitted_values` runs, so `fit(max_horizon=24, fitted=True)` with 3 models keeps 72 copies of the encoded feature frame alive at once. Streaming the fitted predictions at fit time would avoid retaining `X` at all — worth considering if this is expected to run on wide feature sets.

mlforecast/feature_encoders.py
```diff
@@ -0,0 +1,338 @@
...
+                        / (pl.col("__prior_count").fill_null(0) + self.smoothing)
+                    )
+                    .fill_nan(None)
+                    .fill_null(self.prior)
+                    .cast(pl.Float32)
+                    .alias(f"{column}__mean")
+                )
+                .select(column, "__encoder_time", f"{column}__mean")
+            )
+            encoded = encoded.join(stats, on=[column, "__encoder_time"], how="left")
```
nasaul, line 301: **Null categories get `null` here but `global_mean_` in `transform`.** This join uses polars' default `nulls_equal=False`, and unlike `transform` (line 333) there's no `fill_null` afterwards.

With `X = pl.DataFrame({"category": ["a", None, "a", None]})`, `fit_transform` yields `category__mean = [0.0, null, 1.0, null]` while `transform` on the same frame yields `[2.0, 2.5, 2.0, 2.5]`. `LinearRegression` and most sklearn regressors then error during `fit`; tree models happily train on NaN and meet a finite value at predict time.

Adding the same `fill_null(self.prior)` / `fill_null(global_mean_)` after the join should do it. Covered by `test_target_encoder_null_category_is_encoded_at_fit_time`.

mlforecast/feature_encoders.py
```diff
@@ -0,0 +1,338 @@
...
+
+class PolarsOrdinalEncoder(_PolarsEncoder):
+    """Ordinal encoder for Polars dataframes; unseen categories map to ``-1``."""
+
+    def fit_transform(self, X: pl_DataFrame, _y: np.ndarray) -> pl_DataFrame:
+        self._validate_X(X)
+        self.mappings_ = {}
+        for column in self.columns:
+            encoded_name = f"{column}__ordinal"
+            values = X.get_column(column).unique(maintain_order=True)
```
nasaul, line 118: `unique(maintain_order=True)` includes `null`, so it consumes a code from `np.arange(len(values))` — but the left join below uses `nulls_equal=False` and can never match it, so those rows fall through to `fill_null(-1)` anyway.

Net effect: nulls are encoded as "unseen", and the code space is left with a hole. `["a", None, "b"]` encodes as `[0, -1, 2]` — `"b"` gets `2` instead of `1`. Filtering nulls out of `values` (or joining with `nulls_equal=True`) fixes it.

Covered by `test_ordinal_encoder_does_not_spend_a_code_on_null`.

mlforecast/feature_encoders.py
```diff
@@ -0,0 +1,338 @@
...
+                {column: values, encoded_name: np.arange(len(values), dtype=np.int32)}
+            )
+        return self.transform(X)
+
+    def transform(self, X: pl_DataFrame) -> pl_DataFrame:
+        self._validate_X(X)
+        out = X
+        for column, mapping in self.mappings_.items():
+            encoded_name = mapping.columns[-1]
+            out = out.join(mapping, on=column, how="left").with_columns(
```
nasaul, line 129: `PolarsTargetEncoder` deliberately brackets its joins with `with_row_index("__encoder_row")` + `sort`, but this one (and `PolarsCountEncoder.transform`) doesn't, and `pl.DataFrame.join` defaults to `maintain_order=None`, for which polars explicitly does not guarantee output ordering.

Order happened to hold on polars 1.31 in a 200k-row check, so this isn't currently broken — but if a future or streaming engine reorders, the encoded column silently misaligns with `y`, which is the worst failure mode to debug. `maintain_order="left"` here and in the count encoder would make it consistent with the target encoder.

No test for this one — I couldn't produce a deterministic repro.

mlforecast/core.py
```diff
@@ -1179,11 +1179,14 @@ def _transform_per_horizon(
 
             X_h = ufp.filter_with_mask(X_h, valid)
             y_h = y_h[valid]
+            context = {
```
nasaul, line 1182: Minor, perf: this dict is built unconditionally, so `ufp.filter_with_mask(prep[self.time_col], valid).to_numpy()` runs for every horizon of every model — including the majority case of no `feature_encoders` at all. Since `generator_factory()` is re-created per model, `max_horizon=100` with 3 models means 300 extra filtered copies of the time column per fit. Building it lazily would keep the cost on the users who asked for it.

nasaul (commented):
Nice work — the causal target encoder and its leakage test are the right shape, and `test_target_encoder_cross_validation_is_leakage_free` is a genuinely good test.

I reviewed this end to end and reproduced eleven issues locally. Seven are deterministic, so rather than describe them I pushed **`tests/test_feature_encoders_review.py`** to this branch (`f89dd93`) with one failing test each. They're red right now — that's intentional, they're the acceptance criteria. Each derives its expectation from an independent source (the same data in a different row order, an earlier CV fold, hand arithmetic) rather than a constant, so they stay valid however you choose to fix each one. Your existing nine tests still pass untouched.

Happy to push fixes too if you'd rather — just say the word.

### The three I'd treat as blocking

1. **`_accepts_context` breaks every stock sklearn transformer.** `TransformerMixin.fit_transform` is `(X, y=None, **fit_params)`, so the `VAR_KEYWORD` clause matches and sklearn forwards `context` into `fit()`. `feature_encoders=[StandardScaler()]` raises `TypeError` — and that's exactly the interface the new docstring advertises.
2. **Fitted values are misaligned** whenever the input isn't already sorted by id/time, in both the recursive and direct paths. With rows interleaved `b,a,b,a,…` the fitted values come back as `[0, 0, 6, 1, 6.5, 1.5, 7, 2]` where the correct answer is `[0, 1, 1.5, 2, 0, 6, 6.5, 7]` — the right numbers, permuted. Nothing raises.
3. **Target leakage in CV after the first fold.** `cross_validation(fitted=True, refit=False)`: fold 0 is causal `[0, 1, 1.5, 2, 2.5]`, fold 1 is a flat `3.0` — the global mean of the whole training target.

### Two more that have no line to hang on

**`save`/`load` drops `feature_encoders` (`forecast.py:2160`).** `load` rebuilds with `MLForecast(models=models, freq=ts.freq)`, so `fcst.feature_encoders == []`. Prediction still works because the encoders live inside `_EncodedModel`, but `MLForecast.load(path).fit(df)` retrains with no encoding at all — and it doesn't error, because `_EncodedModel.__getattr__` forwards `__sklearn_clone__` to the wrapped estimator, so `clone()` silently unwraps it back to a bare `DummyRegressor`. Covered by `test_save_load_refit_keeps_encoders`.

**`fit_models`' documented contract changed (`forecast.py:627`).** It's public API and the docstring still says the factory yields `(h, X_h, y_h)`, but line 687 now unpacks four values. Any external caller supplying its own factory gets `ValueError: not enough values to unpack`. The two new parameters (`encoder_context`, `store_fitted_X`) are undocumented as well.

The remaining findings are inline below.

jmoralez:
@janrth the reason these can't be sklearn transformers is due to the target transformer, right? I'd prefer if they were so they can be used in pipelines and keep the flexibility of having different preprocessing steps for different models (I understand this will be inefficient for direct forecasting but the mappings stored shouldn't be too heavy) while also preventing introducing even more arguments to the MLForecast constructor.

Also the target transformer looks at the future which we try very hard to avoid in the feature generation, have you had good results using it?

janrth:
> @janrth the reason these can't be sklearn transformers is due to the target transformer, right? I'd prefer if they were so they can be used in pipelines and keep the flexibility of having different preprocessing steps for different models (I understand this will be inefficient for direct forecasting but the mappings stored shouldn't be too heavy) while also preventing introducing even more arguments to the MLForecast constructor.
> 
> Also the target transformer looks at the future which we try very hard to avoid in the feature generation, have you had good results using it?

For sure, let us think about the pipeline topic together.

On the data leakage, I think the current target encoder is doing what the target encoder from category_encoder is doing. It all depends on the right usage when using tuning or cross validation. As long as we do not use some random fold cv method that would use future data, then we are fine. The encoder currently sees all the training data, which to me is not data leakage.

Some people also want to avoid in-training leakage and try to follow some more advanced structure like catboost is doing (some kind of rolling or moving window method inside the training data). I think both methods are fine and the idea is also to add more encoders. I just wanted to start with some common and basic encoders. Personally, I have used the regular target encoder (in combination with count encoder and found good results.).

Back to the pipeline topic: I do not believe that category_encoder are based on pandas and here I tried to use pl. I think sklearn can accept pl in a pipeline and therefore could be compatible with the pl encoders here. Below is some example that would work, but let us think about that and find examples that maybe do not work:

```
from sklearn.pipeline import make_pipeline
from sklearn.preprocessing import StandardScaler

import numpy as np
import polars as pl

from mlforecast.feature_encoders import PolarsTargetEncoder

pipeline_data = pl.DataFrame({
    "unique_id": np.repeat(["a", "b"], 6),
    "ds": list(range(6)) * 2,
    "y": np.arange(12, dtype=float),
    "category": np.repeat(["A", "B"], 6),
}).with_columns(pl.col("category").cast(pl.Categorical))

sklearn_pipeline = make_pipeline(StandardScaler(), LinearRegression())
pipeline_fcst = MLForecast(
    models=sklearn_pipeline,
    freq=1,
    lags=[1],
    feature_encoders=[PolarsTargetEncoder(["category"], drop_original=True)],
)
pipeline_fcst.fit(pipeline_data, static_features=["category"])
pipeline_fcst.predict(1)
```

jmoralez:
What I meant is: if these transforms are just for speed, why can't we have them as regular sklearn transformers? e.g.

```python
import numpy as np
import polars as pl
from sklearn.base import BaseEstimator, TransformerMixin
from sklearn.utils.validation import check_is_fitted

class PolarsCountEncoder(BaseEstimator, TransformerMixin):
    def __init__(
        self,
        cols: list[str],
        unseen_value: float = 0,
    ):
        self.cols = cols
        self.unseen_value = unseen_value
    
    def fit(self, X: pl.DataFrame, y=None) -> "CountEncoder":
        missing = [col for col in self.cols if col not in X.columns]
        if missing:
            raise ValueError(f"The following columns are missing from X: {missing}")
        self.counts_: dict[str, pl.DataFrame] = {
            col: X.get_column(col).value_counts(normalize=False)
            for col in self.cols
        }
        self.feature_names_in_ = np.array(X.columns)
        return self
    
    def transform(self, X: pl.DataFrame) -> pl.DataFrame:
        check_is_fitted(self)
        return X.with_columns(
            pl.col(col).replace_strict(
                counts.get_column(col),
                counts.get_column("count"),
                default=self.unseen_value,
                return_dtype=pl.UInt32,
            )
            for col, counts in self.counts_.items()
        )

    def get_feature_names_out(self, input_features=None) -> np.ndarray:
        check_is_fitted(self)
        return self.feature_names_in_.copy()
```
that allows us to use them in scikit-learn pipelines without any changes to the mlforecast code

PR #719, opened 2026-08-19
feat: add preprocessing, fit, and prediction reports

janrth:
## Summary

Adds lightweight, inspectable reports to show what happens at MLForecast's preprocessing and model-execution boundaries.

- `feature_preparation_report_` records input/output backend, shape, dtypes, memory, row drops, MLForecast-owned conversions, and preprocessing duration.
- `model_fit_report_` records estimator fit-call count and elapsed time per configured model, including direct-model horizon fits.
- `fit_report_` records end-to-end `MLForecast.fit()` elapsed time and process RSS change, with the nested model-fit report.
- `predict_report_` records end-to-end `MLForecast.predict()` elapsed time, process RSS change, and requested horizon.

The reports are created after successful operations and provide both dataclass attributes and `to_dict()` for logging/benchmarking.

## Scope and interpretation

RSS measurements use `psutil` when available. They measure process-level resident-memory change at the operation boundary, so they are useful for comparing configurations such as pandas vs. Polars or `as_numpy=True` vs. `False`. They do not attribute memory to a specific native LightGBM/XGBoost/CatBoost allocation, nor do they represent temporary peak memory that was released before the measurement ended.

Estimator timing is measured around the actual cloned estimator `.fit()` call. It therefore captures model-side work and MLForecast conversions that happen immediately at that boundary, while `fit_report_` retains the complete MLForecast fit duration.


nasaul (commented):
Thanks for this — the reports are a nice, lightweight way to see what MLForecast is doing at each boundary. A few things I think are worth addressing before merge:

## 1. Make the instrumentation opt-in

`forecast.py`'s `preprocess()`, `fit_models()`, `fit()`, and `predict()` all run this unconditionally now — there's no constructor flag, env var, or branch to skip it. Every `preprocess()` call runs `df.memory_usage(index=True, deep=True)` / `pl_DataFrame.estimated_size()`, and every `fit_models()`/`fit()`/`predict()` call does two `psutil.Process().memory_info()` reads, whether or not anyone ever reads `*_report_`.

For a single `.fit()` this is negligible, but MLForecast gets used in loops where it isn't — `cross_validation` with many windows/refits, hyperparameter search, per-series fitting — and the `deep=True` memory scan in particular isn't free on wide frames with object columns.

Could this be opt-in? e.g.:
```python
MLForecast(..., report_level="off")       # default: no instrumentation, current pre-PR behavior
MLForecast(..., report_level="basic")     # today's timing + shape info
MLForecast(..., report_level="detailed")  # + RSS / deep memory scans
```
That way users who don't care pay nothing, and the pricier memory bookkeeping is only paid for when someone actually asks for it.

## 2. Add a report-to-report comparison method

The most common thing someone will want to do with these reports is compare two of them — "did this get faster/lighter after I changed X" (a config, a dependency bump, `as_numpy`, etc.). Right now that means manually pulling fields off two report objects and diffing them by hand.

Could we add a `.diff(other)` (or a shared comparison helper across the four dataclasses) that returns a structured comparison — delta, delta %, and a flag when it crosses some threshold? PR #725's `tests/mem_compare.py` already implements almost exactly this logic for the CI memory harness (base vs. PR, %-and-absolute-MB threshold, a "noisy run" flag). Worth factoring that comparison logic into something shared rather than having two independent implementations of "compare two performance snapshots" in the codebase.

## 3. These reports only cover compute cost, not forecast quality

`model_fit_report_` / `fit_report_` are exclusively about computational cost (fit time, fit calls, RSS). For a forecasting library, "how did this run perform" is ambiguous between *how expensive it was* and *how accurate it was*, and right now these reports only ever answer the first question. Naming them `*_report_` without qualification risks someone assuming they'll find accuracy info here.

When multiple models are configured, it'd be valuable to also report a per-model accuracy comparison — reusing the idea PR #725's `bench_pipeline.py` already applies inline (`smape` across models on held-out data) rather than inventing new evaluation code. Concretely: when `fitted=True`, or after `cross_validation()`, compute per-model error metrics and attach them alongside `model_fit_report_` — e.g. a `model_metrics_` dict of `{model_name: {metric_name: value}}` — so a user can see cost and accuracy side by side. This could be opt-in too, since it needs actuals and so only applies to the fitted-values/cross-validation paths, not a plain `predict()` on the future.

## 4. `cross_validation()` silently overwrites reports per fold

`cross_validation()` calls `self.fit()` once per fold that refits (`forecast.py`, the `for i_window, (cutoffs, train, valid) in enumerate(splits):` loop, `self.fit(train, ...)` at `should_fit`). Since `fit_report_` / `feature_preparation_report_` / `model_fit_report_` are plain instance attributes, each fold's call overwrites whatever the previous fold wrote. With the default `refit=True` and, say, `n_windows=5`, after `cross_validation()` returns, `fcst.fit_report_` reflects **only the last fold** — the other four folds' timing/memory are silently discarded, with no indication in the API that this happened.

This is the same overwrite pattern already solved for prediction-interval calibration internally, where `_calibration_fit_reports` accumulates a list of per-fold `ModelFitReport`s specifically to avoid this problem. `cross_validation()` should probably get the same treatment — accumulate a report per fold (e.g. `cv_fit_reports_: List[FitReport]`) rather than leaving only the final fold's numbers behind.

mlforecast/forecast.py
```diff
@@ -1171,6 +1262,13 @@ def fit(
         Returns:
             MLForecast: Forecast object with series values and trained models.
         """
+        fit_start = perf_counter() if self.report_level != "off" else None
+        fit_rss_start = (
+            get_process_rss_bytes() if self.report_level == "detailed" else None
+        )
+        has_calibration = prediction_intervals is not None
+        if has_calibration and self.report_level != "off":
+            self._calibration_fit_reports: List[ModelFitReport] = []
```
nasaul: **medium** — This list is created before `_conformity_scores` runs but only deleted on the success path (line 1434). If the intervals fit raises — too-few-samples is a common user-facing error — the attribute survives, and every later plain `fit()`/`fit_models()` appends to it through the `hasattr` check at line 757, growing unboundedly and keeping report objects alive. A later intervals fit then silently resets it.

Reproduced: after one failed intervals fit, two clean `fit()` calls leave `len(_calibration_fit_reports) == 2`.

`try/finally` around the calibration block, or pop the attribute in the non-calibration branch.

mlforecast/forecast.py
```diff
@@ -1851,6 +1984,17 @@ def predict(
                         forecasts = _rescale_interval_columns(
                             forecasts, list(model_names), level_, sigma_tgt
                         )
+            if predict_start is not None:
+                self.predict_report_ = PredictReport(
```
nasaul: **low/medium** — `predict_report_` is written by *every* internal predict, not just user-facing ones. `fit(df, prediction_intervals=PredictionIntervals(n_windows=2, h=2))` leaves `predict_report_` populated with `horizon=2` even though the user never called `predict`; `cross_validation` leaves it reflecting whichever fold ran last. Reading `fcst.predict_report_` after a fit gives you a report for work you didn't ask for. A re-entrancy flag, or expose the CV/calibration predicts under their own attribute.

mlforecast/forecast.py
```diff
@@ -1926,6 +2070,18 @@ def cross_validation(
         results = []
         cv_models = []
         cv_fitted_values = []
+        for attribute in (
```
nasaul: **low** — This loop clears `cv_model_metrics_*` unconditionally, but `cv_fit_reports_` / `cv_feature_preparation_reports_` / `cv_model_fit_reports_` / `cv_report_folds_` are only *initialized* further down when `report_level != "off"`. So: run CV with reports on, set `report_level = "off"`, run CV again → the previous run's report lists are still on the object and look current. Adding those four names to this tuple fixes it.

mlforecast/model_report.py
```diff
@@ -0,0 +1,391 @@
...
+        )
+        evaluated_nw = nw.from_native(evaluated, eager_only=True)
+        for model in models:
+            summary[model][name] = float(evaluated_nw[model].mean())
+    return summary
+
+
+def metric_name(metric: Callable) -> str:
+    """Return the stable display name used for a metric callable."""
+    return getattr(metric, "__name__", type(metric).__name__)
```
nasaul: **low/medium** — `type(metric).__name__` is `"partial"` for every `functools.partial`, which is the idiomatic way to pass a parameterized loss. `model_metrics=[partial(mase, seasonality=7), partial(mase, seasonality=12)]` raises `ValueError: model_metrics must have unique names.` at construction (reproduced), and a single partial gets keyed under the useless name `"partial"`. Unwrap `partial.func.__name__`, or accept `(name, callable)` pairs.

mlforecast/model_report.py
```diff
@@ -0,0 +1,391 @@
...
+        return None
+    model_seconds: Dict[str, float] = {}
+    for report in reports:
+        for name, seconds in report.model_seconds.items():
+            model_seconds[name] = model_seconds.get(name, 0.0) + seconds
+    return ModelFitReport(
+        elapsed_seconds=sum(report.elapsed_seconds for report in reports),
+        fit_calls=sum(report.fit_calls for report in reports),
+        model_seconds=model_seconds,
+        rss_start_bytes=reports[0].rss_start_bytes,
```
nasaul: **low** — `rss_start_bytes` comes from `reports[0]` and `rss_end_bytes` from `reports[-1]`, so the aggregate `rss_delta_bytes` spans everything that happened *between* the fits — all of the calibration CV's preprocessing and prediction — rather than the fits themselves. It can even come out negative while both component deltas were positive. Either sum the per-report deltas or leave the aggregate's RSS fields `None`; every other field here is a true sum, so this one reads as if it were too.

mlforecast/model_report.py
```diff
@@ -0,0 +1,391 @@
...
+@dataclass
+class FitReport(_ComparableReport):
+    """Observed end-to-end cost of a successful ``MLForecast.fit`` call."""
+
+    elapsed_seconds: float
+    rss_start_bytes: Optional[int]
+    rss_end_bytes: Optional[int]
+    model_fit_report: Optional[ModelFitReport]
+    calibration_model_fit_report: Optional[ModelFitReport]
+    final_model_fit_report: Optional[ModelFitReport]
```
nasaul: **low** — `_numeric_fields` walks all three of these generically, and by construction `model_fit_report` is the aggregate of `calibration_model_fit_report` + `final_model_fit_report`. So `FitReport.diff()` emits the same seconds and `fit_calls` two or three times under different keys, and a single regression counts two-to-three times against the thresholds. Skip the aggregate (or the two components) when flattening.

mlforecast/model_report.py
```diff
@@ -0,0 +1,391 @@
...
+@dataclass
+class FeaturePreparationReport(_ComparableReport):
+    """Observed input and output characteristics of a preprocessing call.
+
+    The report describes only operations performed by MLForecast.
+    """
+
+    input_backend: str
+    output_backend: str
+    input_shape: Tuple[int, int]
```
nasaul: **low** — `input_shape`/`output_shape` are tuples, which `_numeric_fields` drops, so `diff()` can never flag a row- or column-count change between runs — arguably the most useful cross-run comparison there is. Flattening them as `input_shape.rows`/`input_shape.cols` would cost a couple of lines.

mlforecast/model_report.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+def _memory_bytes(data: Any) -> int:
+    if isinstance(data, np.ndarray):
+        return int(data.nbytes)
+    if isinstance(data, pd.DataFrame):
+        return int(data.memory_usage(index=True, deep=True).sum())
+    if isinstance(data, pl_DataFrame):
+        return int(data.estimated_size())
+    return 0
```
nasaul: **low** — Returning `0` for an unrecognized input means an unknown memory footprint is reported as a definite zero bytes, so memory ratios and diffs come out wrong rather than absent. `Optional[int]` returning `None` here matches how the RSS fields already handle "unavailable".

nasaul (commented):
Reviewed the diff on `mlforecast/forecast.py`, `mlforecast/model_report.py`, and `tests/test_model_report.py`. All 24 new tests pass, and the save/restore of state around `_conformity_scores` is sound and correctly scoped.

Findings below: 3 state-lifecycle bugs in `forecast.py` (one reproduced) and 5 correctness issues in the report/diff math. Findings 6 and 7 both live in the `diff`/`ReportComparison` machinery, which has no in-repo consumer yet — dropping that until something actually compares reports would remove them along with a chunk of surface area.

Two smaller things not filed inline: `report_level="detailed"` calls `memory_usage(deep=True)` on the raw input, an O(n) scan over every object column that perturbs the very timing it's measuring; and enabling `model_metrics` makes the internal calibration CV inside `fit(prediction_intervals=...)` compute `cv_model_metrics_*` and immediately discard it.

jmoralez:
Thanks for this, and for iterating on @nasaul's comments. I agree with the goal: being able to see where time and memory go at MLForecast's boundaries (preprocess, each estimator fit, predict, CV windows) is useful, and the boundaries you picked are the right ones. My concern is with the mechanism, and I'd like to propose a different shape before this grows further.

## Concern

The reporting is implemented inline: every boundary now carries `if self.report_level != "off"` branches, timers and RSS reads, and the results are stored as attributes on `MLForecast` (`fit_report_`, `model_fit_report_`, `predict_report_`, `cv_*_reports_`, ...). Because these methods call each other recursively (`fit(prediction_intervals=...)` → `cross_validation` → `fit` → `fit_models`, and `cross_validation` → `predict`), the attributes overwrite each other, which is what forced `_calibration_fit_reports`, `aggregate_model_fit_reports`, the `_CV_STATE_ATTRIBUTES` snapshot/restore in `_conformity_scores`, and the per-fold `cv_*` lists. Most of the inline findings in the last review (the list that leaks when the intervals fit raises, `predict_report_` being written by internal predicts, stale `cv_*` lists, the aggregate RSS spanning the gaps between fits, double counting in `diff()`) are symptoms of that one choice rather than separate bugs. The net result is +220 lines in `forecast.py` and a 390-line module that every future change to `fit`/`predict`/`cross_validation` has to keep consistent.

## Proposal: let `MLForecast` announce its stages, and do the bookkeeping in a callback

`MLForecast` gets a `callbacks` list and emits start/end/error events for its stages. All reporting logic lives in a `Callback` subclass, so the general path stays almost untouched and nothing report-related is stored on the forecaster.

```python
# mlforecast/callbacks.py
class Callback:
    def on_start(self, stage, fcst, **kwargs): ...          # kwargs: bound arguments of the call
    def on_end(self, stage, fcst, result, **kwargs): ...
    def on_error(self, stage, fcst, error, **kwargs): ...   # the error is re-raised afterwards
```

Stages: `preprocess`, `fit`, `fit_models`, `predict`, `cross_validation` (kwargs are the method's bound arguments with defaults applied), `fit_model` (`name`, `model`, `X`, `y`, `h`, `fit_kwargs`; one event per estimator `.fit`, with `X`/`y` exactly as the estimator receives them) and `cv_window` (`i_window`, `cutoffs`, and the window's result frame on end).

On the `forecast.py` side this is a decorator on the public methods and two emit calls in the CV loop; the method bodies don't change:

```python
@_stage("fit")            # likewise on preprocess, fit_models, p
def fit(self, df, ...):
    ...

@_stage("fit_model")
def _fit_model(self, name, model, X, y, h, fit_kwargs):
    return clone(model).fit(X, y, **fit_kwargs)

for i_window, (cutoffs, train, valid) in enumerate(splits):
    _emit(self, "on_start", "cv_window", i_window=i_window, cutof
    ...
    _emit(self, "on_end", "cv_window", result=result, i_window=i_window, cutoffs=cutoffs)
```

`_stage` is ~20 lines: `if not self.callbacks: return method(...)`, otherwise bind the arguments (signature cached at import), emit start, run, emit end or error. With no callbacks the cost per call is one attribute check, so there's no need for a `report_level` switch.

The user-facing part becomes a `Profiler` callback:

```python
profiler = Profiler(memory=True)   # memory=False skips the psuti
fcst = MLForecast(models=..., freq="D", lags=[1], callbacks=[profiler])
fcst.cross_validation(df, n_windows=3, h=7)
profiler.spans     # tree: cross_validation -> cv_window -> fit -> preprocess, fit_models -> fit_model ...                                               profiler.to_df()   # one row per stage: stage, depth, seconds, rsndow, rows
profiler.to_df().query("stage == 'fit_model'").groupby("model")["seconds"].sum()
```

Because the profiler sees the nesting, the hard parts of the currh intervals shows the calibration `cross_validation` as a childnext to the final `preprocess`/`fit_models` (no aggregation code)thing is overwritten), an internal `predict` is a child of`cross_validation` rather than a top-level report, and there is npshot or clear.

This also covers the review asks: opt-in is "no callbacks", the mory=True)`, comparing two runs is comparing two DataFrames from`to_df()` (no bespoke `diff()`), and per-model accuracy is already `utilsforecast.evaluation.evaluate(cv_df, metrics=[smape])` on the frame `cross_validation`
returns; per-window scoring is a few-line callback on `cv_window` rather not add a second metrics implementation to the library.

## Concretely, what I'd like this PR to become

- Add `callbacks: Optional[List[Callback]] = None` to `MLForecastrs, and the `Callback` + `Profiler` classes in`mlforecast/callbacks.py` (next to `SaveFeatures`).
- Drop `report_level`, `model_metrics`, the four report dataclassn`, `aggregate_model_fit_reports`, `summarize_model_metrics`, the`cv_*_reports_`/`cv_model_metrics_*` attributes and the `_CV_STATE_ATTRIBUTES` save/restore.
- Tests along the lines of: stages and bound arguments for `fit` event per horizon for direct models, calibration nested inside`fit`, every CV window recorded, error unwinding, and a small custom callback.
I put together a prototype of the above to check it holds up: theines, the existing suite passes unchanged, and it's on [feat/callbacks-proposal](https://github.com/Nixtla/mlforecast/tree/feat/callbacks-proposal) if you want to start from it or compare.

janrth:
@jmoralez I think the latest changes are pretty close to your branch. I also removed the cv model performance part, following your suggestion.

jmoralez:
Let's wait for #739 because it'll create conflicts.

jmoralez (commented):
Thanks for reworking this — the callback shape is right and this is a big improvement over the inline version. `forecast.py` gains five decorator lines plus two `_emit` calls, nothing report-related is stored on the forecaster, and `_calibration_fit_reports` / `aggregate_model_fit_reports` / `_CV_STATE_ATTRIBUTES` are all gone. I stress-tested span balance across `fit(prediction_intervals=...)`, `cross_validation(prediction_intervals=..., level=...)`, `cross_validation(fitted=True, refit=False)`, `max_horizon`, `predict(new_df=...)` and two failure paths — the profiler's stack came back empty every time. It also merges cleanly with #739 and the suite passes (358 passed, 2 skipped).

**1. Move these out of `mlforecast/callbacks.py` and rename the parameter.** That module is the home of `SaveFeatures`, a plain callable for `predict(before_predict_callback=...)`, and `*_callback` is the established spelling for that concept across `forecast.py`, `lgb_cv.py` and `distributed/forecast.py`. Putting an observer protocol next to it means one API doc page has to cover two unrelated things and `from mlforecast.callbacks import Callback, SaveFeatures` reads as if they're siblings. A new module (`mlforecast/hooks.py` with `Hook`/`hooks=`, or `stage_callbacks.py` with `StageCallback`/`stage_callbacks=` — keep module, class and parameter consistent) leaves `callbacks.py` and `docs/callbacks.html.md` untouched and drops this PR's footprint on existing files to `forecast.py` alone. Nothing has shipped, so there's no deprecation cost.

**2. The wrapper frame breaks `stacklevel` attribution, on the no-callbacks fast path.** `callbacks.py:80` is `return method(*args, **kwargs)`, so the extra frame is present for every user:

```
$ pytest tests/test_transfer_conformal.py::test_frozen_backtest_uses_source_model
# main:     tests/test_transfer_conformal.py:409: UserWarning: Transfer recalibrate level=90 ...
# this PR:  mlforecast/callbacks.py:80:          UserWarning: Transfer recalibrate level=90 ...
```

Any `warnings.warn(..., stacklevel=N)` whose chain crosses `preprocess`/`fit`/`fit_models`/`predict`/`cross_validation` now points inside mlforecast instead of at the caller. It's a short list — `forecast.py:1732`, `forecast.py:292`, `conformal_prediction.py:404` and `:412`, plus the three `stacklevel=4` sites in `conformal_prediction.py` worth checking — so bumping those and adding a regression test is probably the right fix. The reported line differs between the two paths (`:80` without callbacks, `:87` with) but the frame count is the same either way, so a uniform `+1` is correct.

**3. Instances that skipped `__init__` raise.** `if not self.callbacks` is unguarded: `pickle.loads(<blob from an older version>).predict(2)` → `AttributeError: 'MLForecast' object has no attribute 'callbacks'`. `save`/`load` is fine (it rebuilds through `_from_ts` → `__init__`), so this is only cross-version direct pickles — but it's a one-liner and matches the defensive style already in the file (`getattr(self, "cv_fitted_values_", [])`). A class-level default, or `getattr(self, ..., ())` in the wrapper.

**4. `cv_window` is a second-class stage.** It's the only one emitted by hand and the only one with no error hook, which is what forces `while self._stack.pop()[0] != stage` at `callbacks.py:189`. That loop is doing more work than it looks: it's also cleaning up after `callbacks.py:85`, where the `on_start` emit sits outside the `try`, so a callback raising in `on_start` leaves a stale entry that only gets swept by an enclosing stage's `on_error`. If either is "simplified" without the other, the stack corrupts; and if `stage` is ever absent from the stack, the `IndexError` replaces the user's real exception. Wrapping the CV loop body in `try/except` and emitting `on_error` for `cv_window` fixes the asymmetry, lets this become a guarded pop, and spares every third-party callback from reimplementing the unwinding.

**5. No validation on the constructor argument.** Passing the wrong kind of object fails deep inside `fit` with `AttributeError: '...' object has no attribute 'on_start'`. The rename in (1) makes the `SaveFeatures` mix-up much less likely, but an `isinstance` check is still two lines and turns a confusing failure into a clear one.

**Smaller:**

- `to_df()` on an empty profiler returns a 0×0 frame, so the idiom the class docstring recommends raises `UndefinedVariableError: name 'stage' is not defined`. Construct with the four fixed columns.
- `import psutil` in the hot path; with `memory=True` and psutil absent it surfaces as a bare `ModuleNotFoundError` mid-`fit`. Import in `__init__` with a message naming the argument.
- `h` means two different things in `to_df()`: forecast horizon for `predict`/`cross_validation`, 0-indexed direct-model horizon for `fit_model`. With `max_horizon=3` they land in the same column. Rename one.
- `assert started == stage` is stripped under `-O`, and it's an invariant a user callback can violate.
- `on_error` discards the span entirely, so a failed `fit` leaves no record of how long it ran before failing. Recording it with an error marker seems more useful.

**Maintainability:**

- The stage contract lives in three places — the `@_stage(...)` decorations, the `Callback` docstring, and `_span_details` — with nothing keeping them in sync. `_span_details` indexes `kwargs["df"]`, `kwargs["n_windows"]`, `kwargs["name"]` by hand, so renaming a `cross_validation` parameter breaks the profiler with a `KeyError` at `on_end`. `Recorder` already pins the kwargs for `fit`/`predict`/`fit_model`; extending it to `preprocess`/`cross_validation`/`cv_window` closes the gap cheaply.
- `fit_model` (closure, `forecast.py:678`) and `_fit_model` (method, `:638`) are near-identical names 40 lines apart. Renaming the closure would help. The split itself is right — `_fit_model` gets `X` after the sample-weight column is stripped, so the reported `n_features` is what the estimator actually sees.
- `forecast.py:48` — the import sits at zero indent directly under `if TYPE_CHECKING:`, which reads as if it belongs to it.
- Worth a docstring line each: parent spans' `seconds` include their children (summing over `stage == "fit"` in a CV double-counts); hooks receive live argument objects, so mutating `fit_kwargs` or `df` changes what the estimator gets and is unsupported; `reset()` clears `_stack` too, so it's between-runs-only.
- Since #739, nested calibration stages run on a `_with_ts` scratch copy, so `fcst is my_forecaster` distinguishes user-facing from internal stages — worth documenting, since it's the clean answer to "don't report internal predicts", and since `fcst.models_`/`fcst.ts` are scratch state during those stages.

**Docs:** with the move in (1), `docs/callbacks.html.md` stays as-is and the new module needs its own mkdocstrings page plus an entry in `docs/mintlify/docs.json`. There's a how-to for the predict callbacks (`docs/how-to-guides/predict_callbacks.html`) but none for this; the demo notebook went away in 8f542d4 and nothing replaced it.

PR #720, opened 2026-08-20
fix: support Ray Arrow-backed datasets in LightGBM training

### LINKED ISSUE #713: DistributedMLForecast.fit fails on ray>=2.56: Arrow-backed pandas dtypes rejected by LightGBM
## Summary

Since `ray` 2.56, `DistributedMLForecast.fit` on a Ray Dataset fails because Ray Data now hands LightGBM Arrow-backed pandas columns (`pd.ArrowDtype`), which LightGBM refuses to consume.

CI is currently unblocked with an environment-variable workaround (see below), but the library itself is still broken for any user on `ray>=2.56` with default settings. This issue tracks the real fix.

## What happens

`preprocess` and all feature-equality checks pass; the failure is in `fit`. The user-visible error is unhelpful:

```
RuntimeError: A Ray actor died during training and the maximum number of retries (0) is exhausted.
```

The actual cause is only visible in the Ray worker log for the dead `_RemoteRayLightGBMActor`:

```
File ".../lightgbm/basic.py", line 805, in _check_for_bad_pandas_dtypes
    raise ValueError(
ValueError: pandas dtypes must be int, float or bool.
Fields with bad pandas dtypes: rolling_mean_lag1_window_size7_min_samples1_sub_expanding_mean_lag1: double[pyarrow], rolling_mean_lag1_window_size7_min_samples1_add_expanding_mean_lag1: double[pyarrow], rolling_mean_lag1_window_size7_min_samples1: double[pyarrow]
```

## Root cause

Ray Data 2.56 added `DataContext.enable_arrow_backed_pandas_conversion`, which **defaults to `True`** (`ray/data/context.py`, env override `RAY_DATA_ENABLE_ARROW_BACKED_PANDAS_CONVERSION`). `BlockAccessor.to_pandas` now maps standard Arrow types to `pd.ArrowDtype` instead of numpy dtypes. Ray's own docstring names this exact scenario as the reason to turn it off:

> Set to `False` if pandas UDFs assign multi-dimensional arrays into columns or rely on numpy-only operations (e.g. `%`) that Arrow-backed columns do not implement.

LightGBM's `_check_for_bad_pandas_dtypes` only accepts int/float/bool numpy dtypes, so every engineered feature column is rejected inside the training actor.

The affected code path is the Ray branch of `DistributedMLForecast._fit`, where the materialized dataset is handed to `RayDMatrix` (`mlforecast/distributed/forecast.py`, ~L435):

```python
prep_selected = prep.select_columns(cols=features + [target_col]).materialize()
X = RayDMatrix(prep_selected, label=target_col)
```

## Current workaround (not a fix)

`RAY_DATA_ENABLE_ARROW_BACKED_PANDAS_CONVERSION: 0` was added to the pytest workflow's `env` to restore pre-2.56 dtypes. This only covers CI — it does nothing for users who install mlforecast and call `DistributedMLForecast.fit` themselves.

## Proposed fix

Make the Ray models robust to Arrow-backed dtypes rather than depending on a global Ray flag:

1. Cast Arrow-backed numeric columns to their numpy equivalents before constructing the `RayDMatrix` in `DistributedMLForecast._fit`, so mlforecast works under Ray's default configuration.
2. Alternatively (or additionally), set `enable_arrow_backed_pandas_conversion=False` on the `DataContext` we operate under, so the behaviour is explicit in library code instead of relying on an env var set by the caller.

Option 1 is preferable — it is local to the code that actually needs numpy dtypes and doesn't mutate global Ray state on the user's behalf.

Whichever route is taken, the workaround env var should be removed from `.github/workflows/pytest.yml` in the same PR so CI genuinely covers the supported configuration.

## Related hardening (worth fixing regardless)

This bug cost roughly a day of CI because a one-line dtype error became a 6-hour timeout on every ubuntu job:

- When the LightGBM training actor dies, its placement group is leaked. After the failure `ray.available_resources()` has no `CPU` key at all — the conftest's `num_cpus=2` are still held by `bundle_group_*` reservations — so the *next* test's `ray.data` call blocks forever with `The Ray cluster currently does not have any available CPUs. The Dataset job will hang unless more CPUs are freed up.`
- There is no `pytest-timeout` plugin installed, and `--only-rerun="Socket recv error"` does not match this failure, so nothing fails fast.

Adding a test timeout and tearing down leaked placement groups in the `tests/distributed_ray/conftest.py` session fixture would turn future Ray regressions into a few red minutes instead of a 6-hour hang.

## Reproduction

Verified locally against the versions below (on macOS, by bypassing the Linux-only skip in `tests/distributed_ray/conftest.py`):

- `tests/distributed_ray/test_ray_combine_transforms.py` fails at `dmlf.fit(ray_dataset)` with the traceback above.
- The same test passes with `RAY_DATA_ENABLE_ARROW_BACKED_PANDAS_CONVERSION=0`.

## Versions

| package | version |
| --- | --- |
| ray | 2.56.1 (was 2.55.0) |
| lightgbm_ray | 0.1.9 |
| xgboost_ray | 0.1.19 |
| fugue | 0.9.4 |

Introduced by the ray bump in #703. Only `ubuntu-latest` CI jobs are affected, because that is the only OS that installs `--extra ray`; Python 3.14 is unaffected because the `ray` extra is marker-gated to `python_version < '3.14'`.

kiwoongyoon:
Title: fix: support Ray Arrow-backed datasets in LightGBM training

## Summary

- Temporarily disable Arrow-backed pandas conversion on the preprocessed Ray Dataset while LightGBM-Ray materializes and trains from it.
- Restore the Dataset's original conversion setting even when training raises.
- Remove the CI environment workaround so the existing Ray LightGBM fit coverage runs under Ray's default configuration.

Closes #713

## Test evidence

- Passed a focused source-branch assertion covering successful Ray fitting and context restoration after success and failure.
- Passed workflow assertions confirming the Ray conversion override is absent.
- Passed `python3 -m py_compile mlforecast/distributed/forecast.py` and `git diff --check`.
- The full Ray test could not be run locally because the locked Ray 2.56.1 wheel was unavailable in the offline cache; CI's existing `tests/distributed_ray/test_ray_combine_transforms.py` exercises the real LightGBM-Ray fit without the workaround.

## Risks or notes for maintainers

- The Dataset context setting is temporarily shared with concurrent operations using that same context, but it is restored in `finally` after synchronous training.
- Ray versions that do not expose the setting keep their existing behavior.
- No placement-group cleanup or timeout hardening is included; those are separate from the dtype failure.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -94,8 +94,4 @@ jobs:
       - name: Run pytest
         env:
           RAY_ENABLE_UV_RUN_RUNTIME_ENV: 0
-          # ray >=2.56 converts blocks to pandas with pd.ArrowDtype columns, which
-          # lightgbm rejects ("pandas dtypes must be int, float or bool"). Keep the
-          # pre-2.56 numpy dtypes until the ray models cast the features themselves.
-          RAY_DATA_ENABLE_ARROW_BACKED_PANDAS_CONVERSION: 0
         run: uv run pytest --reruns=2 --reruns-delay=2 --only-rerun="Socket recv error"
--- a/mlforecast/distributed/forecast.py
+++ b/mlforecast/distributed/forecast.py
@@ -432,16 +432,28 @@ class DistributedMLForecast:
                 raise NotImplementedError(
                     "Only spark and dask engines currently support sample weights."
                 )
-            prep_selected = prep.select_columns(
-                cols=features + [target_col]
-            ).materialize()
-            X = RayDMatrix(
-                prep_selected,
-                label=target_col,
+            data_context = prep.context
+            arrow_backed = getattr(
+                data_context, "enable_arrow_backed_pandas_conversion", None
             )
-            for name, model in self.models.items():
-                trained_model = clone(model).fit(X, y=None)
-                self.models_[name] = trained_model.model_
+            if arrow_backed is not None:
+                data_context.enable_arrow_backed_pandas_conversion = False
+            try:
+                prep_selected = prep.select_columns(
+                    cols=features + [target_col]
+                ).materialize()
+                X = RayDMatrix(
+                    prep_selected,
+                    label=target_col,
+                )
+                for name, model in self.models.items():
+                    trained_model = clone(model).fit(X, y=None)
+                    self.models_[name] = trained_model.model_
+            finally:
+                if arrow_backed is not None:
+                    data_context.enable_arrow_backed_pandas_conversion = (
+                        arrow_backed
+                    )
         else:
             raise NotImplementedError(
                 "Only spark, dask, and ray engines are supported."


nasaul:
hello @kiwoongyoon thanks for trying to solve this, as you can see by test hanging this isn't the correct fix, since the first test is failing.

nasaul:
Closing this PR since it has no activity from the author.

PR #722, opened 2026-08-21
fix: support dynamic exogenous features in transfer intervals

### LINKED ISSUE #721: prediction interval in transfer learning with exogenous inputs

I have a time series y(t) that I am training it using samples 0<t<T, but then I want to use it forecasting every day without retraining. So, I am using the transfer learning method as below. I am trying to add prediction interval to my forecast, but i keep getting error about X_df. 

`ValueError: X_df is required for prediction because future values are needed for feature generation or model inputs used during training: ['u'].`

This is a sample code. Any idea on how to fix this issue?

### Versions / Dependencies

mlforecast==1.1.0
python 3.11
MS Windows

### Reproduction script

```
# create a data using simple linear system with input u and output y

T = 20
y = np.zeros(T)
u = np.arange(T,0,-1)

y[0]=0

for i in range(1,T):
    y[i] =-1 * y[i-1] - 2 * u[i]  

df = pd.DataFrame([])
df['y']= y   # output of the system
df['u'] = u   # input of the system

df['ds'] = pd.date_range(start=dt.datetime(2000,1,1), periods=T, freq='D')
df["unique_id"]="A"  # a fake identifier for the data


# Train on the first 10 samples of the data 
fcst = MLForecast(
    models=LinearRegression() ,
    freq='D',   # Daily frequency
    lags= [1]              
)

fcst.fit(    df.iloc[0:10] ,fitted=True, static_features=[] , prediction_intervals=PredictionIntervals(method='conformal_error', n_windows=3))

# modifying the next 5 observations  to be outliers
new_df= df[:].copy().iloc[0:15]
new_df.loc[10:15,'y'] = -100

# use transfer learning to predict the next 5 samples using the modified data and the original model
Y_hat = fcst.predict(h=5, new_df=new_df  , X_df=df[['unique_id','ds','u']].loc[15:19],  transfer_conformal = 'recalibrate', level = [95])
```

### Issue Severity

None

janrth:
## Summary

Fixes #721.

Transfer conformal prediction could not handle dynamic exogenous inputs consistently:

- `recalibrate` and `error_scaled` run a frozen target-domain backtest, but its nested predictions did not receive the validation fold's future dynamic inputs.
- `weighted_conformal` and `scale_aligned_weighted` preprocess target history for density-ratio estimation, but did not preserve the source model's fitted static/dynamic feature configuration.

This PR keeps the fix local to transfer-conformal internals:

- `_frozen_backtest` resolves dynamic model-input columns from the fitted feature schema once, then passes each validation fold's values as `X_df` to its nested prediction.
- The weighted-transfer preprocessing callback preserves the fitted `id_col`, `time_col`, `target_col`, and `static_features` configuration.

Ordinary `predict()` and ordinary `cross_validation()` behavior are unchanged.

## Behavior

All five transfer methods now work with dynamic exogenous features:

- `recalibrate`
- `error_scaled`
- `scale_aligned`
- `weighted_conformal`
- `scale_aligned_weighted`

The added regression test uses a deterministic linear system with a dynamic input and verifies that every method creates interval columns while leaving point forecasts identical to the plain forecast.


mlforecast/forecast.py
```diff
@@ -114,6 +115,16 @@ def _frozen_backtest(
 
     _original_ts = fcst.ts
     try:
+        static_cols = set(fcst.ts.static_features_.columns)
+        dynamic_cols = [
+            col
+            for col in fcst.ts.features_order_
+            if col in new_df.columns and col not in static_cols
```
nasaul: This reimplements `TimeSeries._get_dynamic_exog_cols`, but only excludes static columns. The canonical helper (already used elsewhere for this exact purpose, e.g. `core.py:1889` and `core.py:1923`, paired with `_partition_cols` just like here) also excludes lag-transform output names, date-feature names, and `weight_col`.

As written, if a raw column in `new_df` happens to share a name with a lag/date feature (e.g. a user column literally named `dayofweek` while also using `date_features=["dayofweek"]`), this misclassifies it as a required dynamic column, whereas the canonical helper wouldn't. It also silently diverges from any future changes to the canonical exclusion rules (e.g. `weight_col` handling).

Suggest reusing `fcst.ts._get_dynamic_exog_cols(fcst.ts.features_order_)` (filtered to `new_df.columns`) instead of the inline reimplementation.

mlforecast/forecast.py
```diff
@@ -114,6 +115,16 @@ def _frozen_backtest(
...
+        ]
+        partition_cols = sorted(
+            col for col in fcst.ts._partition_cols if col in new_df.columns
```
nasaul: Every other read of `_partition_cols` in `core.py` (lines 1500, 1891, 1925) goes through `getattr(self, "_partition_cols", ...)`, implying older/loaded `TimeSeries` objects may not always have this attribute. This accesses `fcst.ts._partition_cols` directly, so it would raise `AttributeError` for a source model loaded via `MLForecast.load()` from a save that predates this attribute, if `transfer_conformal` is then used on it.

Suggest matching the existing guard: `getattr(fcst.ts, "_partition_cols", set())`.

nasaul (commented):
Great job, thanks for the solutions, however there are some nits that be solved before merging.

mlforecast/forecast.py
nasaul: Thanks — the `getattr` guard addresses the original nit, but `main` has moved underneath this branch and the fallback now needs to go instead.

`#728` (`perf(pooled): reduce time and memory usage`, 745c5f2) changed `_partition_cols` from an instance attribute assigned in `TimeSeries.fit_transform` (a `set`, built by scanning `self.transforms` for `partition_by`) into a class-level property (`core.py:370`):

```python
@property
def _partition_cols(self) -> List[str]:
    return self._leaf_cols("_pt_cols")
```

It is now derived on demand from the pooled leaves, and partition metadata lives on each leaf as `_pt_cols`, not `partition_by`.

Two consequences for this PR:

1. The legacy fallback in `_frozen_backtest` is dead code after a rebase. A class-level property is always present, so `getattr(fcst.ts, "_partition_cols", None)` can never return `None` — and the fallback reads `getattr(tfm, "partition_by", None)`, which the new engine no longer exposes on leaves. Suggest deleting the fallback and keeping just:

   ```python
   partition_cols = sorted(
       col for col in fcst.ts._partition_cols if col in new_df.columns
   )
   ```

2. The tests that emulate a legacy artifact can't work any more: `delattr(fcst.ts, "_partition_cols")` raises `AttributeError: property '_partition_cols' of 'TimeSeries' object has no deleter`. `test_transfer_recalibrate_supports_legacy_loaded_timeseries` and `test_transfer_recalibrate_supports_legacy_partition_transforms` should be dropped along with the fallback they cover.

Also `test_transfer_conformal_with_dynamic_partition_columns` asserts `fcst.ts._partition_cols == {"promo"}`; the property returns a `list`, so that becomes `== ["promo"]`.

For reference, merging current `origin/main` into 070e358 and running the two conformal test files gives:

```
FAILED tests/test_transfer_conformal.py::test_transfer_conformal_with_dynamic_partition_columns[recalibrate]
FAILED tests/test_transfer_conformal.py::test_transfer_conformal_with_dynamic_partition_columns[error_scaled]
FAILED tests/test_transfer_conformal.py::test_transfer_recalibrate_supports_legacy_loaded_timeseries
FAILED tests/test_transfer_conformal.py::test_transfer_recalibrate_supports_legacy_partition_transforms
4 failed, 94 passed
```

All 98 pass on 070e358 as-is, so this is purely the stale base — a rebase on `main` plus the three changes above should clear it.

nasaul (dismissed):
Approving. The core fix is right: the old code filtered pooled *source* conformity scores by *target* ids (which never match) and set \`n_series = len(ids)\`, guaranteeing a reshape mismatch — all four \`_add_*_transfer_intervals\` paths derive \`n_target\` from \`fcst_df\`, so passing the unfiltered \`_cs_df\` through is correct. The \`_cs_df\`/\`_cs_source_scales_\` class-level defaults also fix \`save()\` raising \`AttributeError\` on a \`preprocess\` + \`fit_models\` instance.

I pushed one commit (a544d0b) rather than leave it as a comment — hope that is alright. \`_frozen_backtest\` was filtering the fit-time dynamic exog and partition columns against \`new_df\`, so a column missing from the target history was silently dropped from \`X_df\`. Because the per-fold \`TimeSeries\` is rebuilt from \`train\` (equally short on columns), \`ts.predict\`'s own \`required_future_cols\` check was empty too, and the failure landed as an opaque feature-name mismatch from inside the model. It now raises naming the columns, mirroring the existing \`X_df is missing future values...\` error in \`core.py\`. One regression test added.

Verified: \`test_transfer_conformal.py\` + \`test_conformal.py\` → 97 passed; \`test_forecast.py\`, \`test_history_warmup.py\`, \`test_conformal_perf_equiv.py\` → 190 passed, 2 skipped. Reverting \`forecast.py\` and \`conformal_prediction.py\` to 745c5f2 fails 9 of the 10 new tests, so the fix is load-bearing. Extra probes pass: all five transfer methods on multi-series pandas and polars with dynamic exog, the same five with \`ids=\` filtering, \`max_horizon\` direct forecasting with exog + date features, and \`TransferConformal(n_windows=)\` override on \`recalibrate\`.

jmoralez:
## Review

Ran the conformal suites against the PR head (`tests/test_conformal.py` + `tests/test_transfer_conformal.py`, 97 passed) plus targeted probes for polars, `ids=` subsets, `max_horizon`, custom column names and user-supplied weights. Two findings look like real bugs; a few more are worth folding in while the branch is open.

### Blocking

**1. User-supplied `TransferConformal.weights` are silently ignored** — `mlforecast/forecast.py:1570`

The same-domain branch is gated on `new_df is None`, and `_weighted_conformal_transfer` never reads `tc.weights`, so `predict(..., new_df=..., transfer_conformal=TransferConformal(method='weighted_conformal', weights=w))` always substitutes auto-fitted DRE weights. Verified empirically: a degenerate `w` (`np.zeros(n_cal); w[0] = 1.0`) produces intervals identical (`np.allclose == True`) to the same call with `weights=None`. The weights pass `__post_init__` validation and are then discarded with no warning.

This PR makes it worse: reworking the `ids`/`is_transfer` block at `forecast.py:1798-1815` removes the old `elif ids is not None -> raise "TransferConformal with DRE weights cannot be used together with ids="` guard for transfer methods — that was the last place the API acknowledged user weights on this path. Either honor `weights` or raise on them.

**2. Pooled groupby key columns missing from the frozen-backtest future columns** — `forecast.py:119`

`_frozen_backtest` forwards only `_partition_cols`, not `_gb_cols`. With e.g. `RollingMean(..., groupby=['grp'], partition_by=['promo'])`, a time-varying `grp`, `static_features=[]` and the default `drop_auxiliary_columns=True`, `core.py:783-802` strips both columns from `features_order_`, so `future_cols` omits `grp`. The new `missing_future` check reports nothing, `X_df` is built without `grp`, and `TimeSeries._update_partition_assignments` (`core.py:1556-1563`) raises the opaque `Partition/group key column(s) ['grp'] not found in X_df or static_features` from inside the backtest loop — precisely the failure this PR set out to name up front. Use `fcst.ts._pooled_aux_cols` (`_gb_cols + _pt_cols`) instead of `_partition_cols`.

### Worth fixing before merge

**3. Dropped `source_cs_df is None` validation** — `mlforecast/conformal_prediction.py:975`

Splitting the `_scale_aligned_transfer` guard lost that condition entirely; it is no longer checked anywhere in the function. Instead of the previous clear `ValueError`, callers now get `TypeError: 'NoneType' object is not subscriptable` deeper in `_apply_scale_alignment` / `_add_conformal_error_intervals`. Unreachable through `MLForecast.predict` today only because `needs_source_cs=True` plus the earlier `self._cs_df is None` check — but the local invariant is no longer enforced.

**4. Schema resolution breaks on the second transfer call** — `forecast.py:118`

`predict` sets `self.ts = new_ts` on success (`forecast.py:1695-1697`), and `new_ts.history_warmup` -> `TimeSeries._fit` rebuilds `features_order_` / `static_features_` from `new_df.columns` (`core.py:780-802`, `736-749`). So a second `predict(new_df=target_b, transfer_conformal=...)` derives `dynamic_cols` / `partition_cols` from **target_a's** schema, not the source fit. If the two targets differ in exogenous columns, or in which columns were inferred static, the new check either raises for a column the source model never needed or silently omits one it does. The PR description's "resolved from the fitted feature schema" holds only for the first transfer call on a given object.

**5. `transfer_preprocess` restores only 4 of the fit-time knobs** — `forecast.py:1647`

`id_col` / `time_col` / `target_col` / `static_features` are forwarded; `weight_col`, `keep_last_n`, `dropna` and `max_horizon` are left at `preprocess`'s defaults. A source fit with `weight_col='w'` stores `_cs_df` features computed with `w` excluded (`core.py:748-749`), but the transfer callback preprocesses with `weight_col=None`, so `w` is retained as an ordinary feature column in `tgt_preprocessed`. Likewise `dropna=True` is assumed even when the source was fit with `dropna=False`, which changes which rows survive and therefore the target sample the density-ratio classifier trains on. Same fix shape as the four already preserved.

### Cleanups (non-blocking)

- `forecast.py:1646` — the isolation fix `copy.deepcopy`s the entire fitted `TimeSeries` (all series data, pooled states, lag-transform buffers, target transforms) on every weighted-transfer predict, purely to give `preprocess` a scratch object it overwrites wholesale via `fit_transform`. On a large panel that roughly doubles peak RSS for the duration of the DRE fit, on top of the `new_ts` copy already built at 1518-1558. Construct a fresh `TimeSeries` from `self.ts`'s config, as the `new_ts` block already does.
- `forecast.py:121` — the required-future-columns set and its error message now exist in three places with four message variants (`core.py:1926-1935`, `core.py:1960-1968`, and here); the existing copies already disagree on ordering and on `getattr(self, '_partition_cols', set())` vs. direct attribute access. Extracting `TimeSeries._required_future_cols()` would also make finding 2 a one-line change.
- `forecast.py:1715` — `cs_ids` does a full `unique()` scan over `_cs_df` (n_windows x n_series x h rows) but is provably dead on every transfer path: its only consumer is `if cs_ids != active_ids and new_df is None:`, and the new `elif not is_transfer` skips the other. Guard the computation.
- `forecast.py:119` — drop the `sorted()`; `_partition_cols` is already deterministic and order-preserving, the following `dict.fromkeys(...)` dedups, and sorting is inconsistent with the adjacent unsorted `dynamic_cols` and with both `core.py` call sites.

### Tests

- `tests/test_transfer_conformal.py:420` — `test_source_score_transfer_supports_target_id_subset` uses the module-scoped `transfer_cp_setup` fixture, and `predict` permanently mutates `mlf.ts`. Inserted at line 411 it now runs before `test_coverage_within_tolerance` and friends, leaving the shared `mlf.ts` in target-domain state (`uids = tgt_*`) while `mlf._cs_df` still holds `src_*` rows. It passes today only because every transfer test re-warms from the same `target_train`; adding a test with a different `new_df`, or a reorder, would give silent cross-test coupling. Use `scope='function'` or re-`fit` inside the test.
- `tests/test_transfer_conformal.py:128` — `test_transfer_conformal_with_dynamic_partition_columns` asserts only `assert_allclose(result['LinearRegression'], baseline['LinearRegression'])`, which is satisfied by the point-forecast column alone. If a regression made `level=[90]` silently produce no `LinearRegression-lo-90` / `-hi-90` columns (e.g. the `if self._cs_df is None: warn` branch at `forecast.py:1699-1704` firing instead of raising), the test would still pass. Mirror the two interval-column asserts from its sibling `test_transfer_conformal_with_dynamic_exog`.

---

https://claude.ai/code/session_01GjEfdeqseLjmg9TzWgJ5f5

janrth:
Status update, since the review thread went quiet after the last fix round.

@jmoralez's review (2026-09-08) flagged 2 blocking bugs, 3 more worth fixing before merge, 4 cleanups, and 2 test gaps. `504d5c1` (2026-09-09) addresses every one of them:

**Blocking**
- **User-supplied `TransferConformal.weights` silently ignored** — `_weighted_conformal_transfer` now honors `tc.weights` (array or callable form), with a length check against the source calibration set; `scale_aligned_weighted` inherits this by delegation. Covered by `test_user_supplied_weights_are_honored`, `test_user_supplied_callable_weights_receive_source_features`, `test_user_supplied_weights_wrong_length_raises`.
- **Pooled groupby key columns missing from frozen-backtest future columns** — new `TimeSeries._required_future_cols()` folds in `_pooled_aux_cols` (groupby + partition leaf columns, statics excluded) and replaces the three copies of this logic that had drifted apart. Covered by `test_transfer_conformal_with_pooled_groupby_columns` and `test_predict_names_missing_pooled_groupby_column`, reproducing the exact scenario from the review.

**Worth fixing before merge**
- **Dropped `source_cs_df is None` validation** in `_scale_aligned_transfer` — restored, with `test_scale_aligned_transfer_requires_source_cs_df`.
- **Schema resolution breaking on a second transfer call** — fixed via `TimeSeries._fitted_static_features` (the fit-time static split as an explicit list, rather than the possibly-`None` inferred one that let a later target re-infer a different split). `test_repeated_transfer_predict_keeps_source_column_schema` asserts two transfer calls in a row on the same object agree with a fresh object's prediction on the same second target.
- **`transfer_preprocess` only forwarding 4 of the fit-time knobs** — `weight_col`/`keep_last_n`/static split are now forwarded too (`test_transfer_preprocess_forwards_fit_time_knobs`). `dropna`/`max_horizon` are deliberately left at their defaults — source calibration rows are model predictions and always fully lagged, so forwarding `dropna=False` would hand the density-ratio classifier NaN target rows; this is called out in the commit message.

**Cleanups** — the whole-`TimeSeries` `deepcopy` (replaced by a config-only `_new_from_config()`), the required-future-columns logic tripled across two files (now the one `_required_future_cols()`), the dead `_cs_df` id scan on paths that never consume it (now lazy), and the unnecessary `sorted()` — all addressed.

**Tests** — the module-scoped fixture that let one test's mutation of `mlf.ts` leak into the next now has a per-test `transfer_cp_isolated` variant; `test_transfer_conformal_with_dynamic_partition_columns` now asserts the interval columns exist, not just the point forecast.

I read the current diff and code directly (not just the commit message) and independently ran `tests/test_transfer_conformal.py` + `tests/test_conformal.py` on the current branch tip (which also has today's merge from `main`): **107 passed**. CI is green across every OS/Python combination.

Given nasaul's earlier approval was auto-dismissed by this fix commit, could one of you take another pass and re-approve if it looks right? Happy to dig further into any specific point first if something still looks off.

janrth:
@jmoralez should we also continue here, or is there something else you have been working on, which might impact the work on this branch, too?

jmoralez:
Let's merge #739 first, that should make this one simpler.

jmoralez:
@janrth can you please try incorporating the changes from main? That'll hopefully reduce the size of the diff and make the changes easier to reason about.

PR #728, opened 2026-09-04
perf(pooled): reduce time and memory usage

jmoralez:
Changes the internal data structures used by pooled lag transformations in order to reduce their runtime and memory usage.


nasaul:
Really nice work — collapsing `_TimestampAggregates` + the per-transform `_impl` hooks into `_CellStore`/`_RowStore` + kernels makes this much easier to reason about, and the `O(n_rows)` fit store is the right call for partitioned panels. Two blocking issues though, both silent-wrong-answer.

**🔴 1. `_RowStore` doesn't filter NaN targets (`pooled.py:486`)**

`_CellStore.__init__` filters (`valid = ~np.isnan(y)`, line 197); `_RowStore.from_rows` doesn't. `main` filtered at the row level in four places (`lag_transforms.py:604,1123,1263,1636`). So any NaN in a window now poisons the whole quantile, and NaN rows count toward `min_samples`.

```
# 2 series x 12 points, target_transforms=[Differences([2])] -> NaN at each series head
# lag_transforms={2: [<tfm>(global_=True)]}, preprocess(dropna=False)
RollingQuantile(p=.5, window_size=4)
  main: [nan nan nan  3.  3.  3.  3.  3.  3.  3. nan nan]
  PR:   [nan nan nan nan nan  3.  3.  3.  3.  3. nan nan]
SeasonalRollingQuantile(p=.5, season_length=3, window_size=2)
  main: [nan nan  3.  3.  3.  3.  3.  3.  3.  3. nan nan]
  PR:   [nan nan  3. nan nan  3.  3.  3.  3.  3. nan nan]
ExpandingQuantile(p=.5)
  main: [nan nan  3.  3.  3.  3.  3.  3.  3.  3. nan nan]
  PR:   all NaN
```

`ExpandingQuantile` is fully dead — its window always reaches the leading NaN cells. `LookupLag` returns the NaN verbatim. Filtering in `from_rows` is the obvious fix, but the counts feeding `min_samples` need to follow.

**🔴 2. `update()` drops a timestamp from pooled `Expanding*` (`pooled.py:1540` / `core.py:511`)**

`append` folds the new column into the channel block but never advances the `Expanding*` accumulator inners — those only move one column per `PooledState.update()`, and coreforecast's `update` folds exactly one value. So every timestamp added via `TimeSeries.update()` / `MLForecast.update()`, plus the last fit ordinal, is silently skipped.

```
# 2 series x 5 points (y=1..5), ExpandingMean(global_=True), then update() with y=100/50
ExpandingMean  main: 15.000000  (12 obs, sum 180)   PR: 17.000000  (10 obs)
ExpandingStd   main: 30.015148                      PR: 32.479364
```

Rolling/seasonal are unaffected (they re-derive from the block) and `EwmK` is fine (`next_src` catches up), so this is specific to the accumulator kernels.

Neither of these is caught by the current suite — worth adding a NaN-target case to the row-kernel tests and an `update()`-then-`predict` case for each pooled accumulator, since both paths are exactly where the new substrate diverges from the old.

**🟠 3. `_ROW_STRIDE` overflow has no guard (`pooled.py:461`)**

The comment states the bound ("a bucket count bounded at 2\*\*23") but nothing enforces it. Past ~8.4M buckets, `bucket * _ROW_STRIDE + ordinal` wraps in int64, `keys` stops being sorted, and every `searchsorted` in `search` returns garbage — silently wrong features, no error. `local` mode multiplies series × distinct partition values, so `LookupLag(partition_by=["holiday_name"])` on a 1M-series panel with 20 holidays is already past it. A `raise` in `from_rows`, or deriving the stride from `ordinal.max() + 1`, would be cheap.

**🟡 4. `_StdKernel` is the moment to fix the variance formula (`pooled.py:749`)**

Not a regression — bit-identical to `main` here. But `main` had this math in three places (`_rolling_std_from_agg`, `_expanding_std_from_agg`, and no seasonal path at all) and this PR collapses them into one `combine()`, so it's the cheapest this fix will ever be. The naive `E[x²] − E[x]²` form loses badly on offset data:

```
# global_ ExpandingStd, y ~ 1e6 + N(0,1), 3 series x 2000 timestamps, vs float128 oracle
#   max rel err 2.192e-03, median 4.059e-04
#   got 0.9899857280434768   expected 0.9921602302145972
```

Three significant figures on a σ≈1 series at 1e6 — sales and price panels look exactly like this. A per-bucket shift (first observation, or a running mean) accumulated into `sumsq` would localize to `_CellStore.__init__` (line 203) plus the `combine`.

---

nasaul (approved):
LGTM

PR #731, opened 2026-09-04
fix: migrate ray models to ray.train

### LINKED ISSUE #713: DistributedMLForecast.fit fails on ray>=2.56: Arrow-backed pandas dtypes rejected by LightGBM

nasaul:
Closes #713.

## Summary

`DistributedMLForecast.fit` has been broken on ray >=2.56 for every user; CI only looked green because the workflow set an env var that users don't have. The fix is to move the ray models off the unmaintained `lightgbm_ray` / `xgboost_ray` onto `ray.train`, because none of the workarounds can be applied from inside `fit()`.

Four things land here:

1. **The fix** — `RayLGBMForecast` / `RayXGBForecast` train via `LightGBMTrainer` / `XGBoostTrainer`. Constructors and `model_` are unchanged.
2. **Hardening** — the placement-group leak that turned this one-line dtype error into a 6-hour CI job is gone, with a `--timeout` and a conftest reclaim as backstops.
3. **ray on python 3.14** — the marker gate is removed, so ubuntu 3.14 now runs the ray suite for the first time.
4. **Cheaper tests and a re-cut xgboost bound** — the ray suite goes from ~14m40s to under 2m, and `xgboost>=2.0.0,<2.1.0` (a `xgboost_ray` artefact) becomes `xgboost>=2.1.4`.

Details below.

## The bug

Since ray 2.56, `Dataset.to_pandas` yields `pd.ArrowDtype` columns. LightGBM's input validation (`_check_for_bad_pandas_dtypes`) whitelists dtypes before handing a buffer to the C++ side and rejects anything it doesn't recognise, so every engineered feature column is refused inside the training actor and `DistributedMLForecast.fit` fails for any user on ray's default configuration.

The user-visible error was useless — `RuntimeError: A Ray actor died during training` — with the real `ValueError: pandas dtypes must be int, float or bool` only visible in the dead actor's worker log.

XGBoost is unaffected: `xgb.DMatrix` accepts arrow-backed pandas natively. This is a LightGBM-only bug.

CI was only green because the workflow set `RAY_DATA_ENABLE_ARROW_BACKED_PANDAS_CONVERSION=0`, which did nothing for actual users.

## Why the fixes proposed in the issue don't work

All three routes are dead ends, which is what drives the approach here:

- **Casting the dataset before `RayDMatrix`** (option 1) doesn't help: the arrow-to-pandas conversion happens *inside* the training actor, not on the driver, so blocks still arrive as `double[pyarrow]`. Forcing pandas blocks via `map_batches(batch_format="pandas")` doesn't change that either.
- **Setting `DataContext.enable_arrow_backed_pandas_conversion = False`** (option 2) doesn't reach `lightgbm_ray`'s actors, which get a fresh default context.
- **The env var** only takes effect when set *before* `ray.init`. mlforecast inside `fit()` always runs after the user's cluster already exists, so it can never be a library-side fix — only advice we'd have to put in the docs.

That leaves no clean way to fix this while staying on `lightgbm_ray`.

## The change

The ray models move to `ray.train`, which is the maintained API (`lightgbm_ray` and `xgboost_ray` are both unmaintained) and, critically, puts the dtype conversion in a train loop we own — where ray's own `normalize_pandas_for_lightgbm` handles it:

- New `RayForecastBase` holds the trainer plumbing; `RayLGBMForecast` / `RayXGBForecast` mix it into `LGBMRegressor` / `XGBRegressor` and build a `LightGBMTrainer` / `XGBoostTrainer` instead of subclassing `RayLGBMRegressor` / `RayXGBRegressor`.
- The train loop builds and fits the library's own estimator in the worker and checkpoints it, as `lightgbm.dask._train_part` does, so `model_` is the estimator that was fitted rather than one rebuilt from native params. That keeps every parameter translation — aliases included — inside the library, and keeps the booster's `evals_result_` and the user's arguments on the way back. The checkpoint carries ray's standard booster artifact alongside it, so `RayTrainReportCallback.get_model` still works on the result.
- LightGBM's network params (`tree_learner="data_parallel"` plus `get_network_params()`) are passed explicitly. Ray's `LightGBMConfig` only stashes them in a per-worker global, so without them each worker trains an independent model on `1/num_workers` of the data and rank 0's is the one checkpointed, silently. XGBoost needs no equivalent: `XGBoostConfig` wraps the loop in a `CommunicatorContext`.
- `_fit`'s ray branch drops `RayDMatrix` and calls `clone(model).fit(prep_selected, target_col=target_col)`.
- `RAY_INSTALLED` now keys off `ray.data` rather than importing `lightgbm_ray`.
- `model_` is still a local `LGBMRegressor` / `XGBRegressor`, so `predict`, `save` and `to_local` are untouched.
- Dependencies: `lightgbm_ray` and `xgboost_ray` out; `ray[data,train]>=2.56` in (2.56 is where `normalize_pandas_for_lightgbm` landed), plus an explicit `lightgbm>=4.6`, which previously arrived only as a `lightgbm_ray` transitive dependency.

### Compatibility

Constructors are unchanged: booster params are still passed as kwargs, with the ray-specific `num_workers` / `resources_per_worker` / `storage_path` keyword-only so they can't collide. Both classes are still `LGBMRegressor` / `XGBRegressor` subclasses, as on `main`, so `isinstance` checks and `clone()` keep working, and the parameter handling is the library's throughout rather than ours.

One thing changes for anyone reaching past the documented surface: `RayParams` is no longer accepted (nothing in the repo or docs ever passed it).

One behavioural change is worth calling out on its own: **the worker count no longer comes from `n_jobs`.** Both old wrappers derived it via `RayParams(num_actors=n_jobs)`, falling back to 1. The default is unchanged, but anyone who set `n_jobs=8` to get 8 actors now gets a single worker with `n_jobs` as its thread count — they want `num_workers=8`.

`resources_per_worker` is the CPU knob: it decides how many CPUs each worker gets and therefore how many threads the booster can use. Left unset it defaults to the cluster's CPUs split evenly across the workers, which is what `xgboost_ray._autodetect_resources` gave the actors, so a default fit keeps the parallelism it had on `main`. `n_jobs` can only lower the count below that share; `model_` carries the requested value rather than the clamp, since it's shipped to the forecasting workers and returned by `to_local`.

`storage_path` is new, and defaults to a temporary directory that's discarded once the fitted model has been read back. Ray's own default would otherwise leave a run under `~/ray_results` per model per fit, which neither old wrapper did; as with ray's default, a local path is single-node, so point it at shared storage on a multi-node cluster.

## Why this bug cost a day of CI, and what stops that recurring

`lightgbm_ray` leaks its placement group when its `fit` raises — on *any* exception, not just this one. The leaked group keeps holding the cluster's CPUs, so the *next* test's `ray.data` call blocks forever rather than failing. That is how a one-line dtype error became a 6-hour job on every ubuntu runner.

Ray Train tears its placement group down on both success and failure, and surfaces the real traceback via `WorkerGroupError`. As backstops:

- CI gets `--timeout=600` (`pytest-timeout` added to the dev group, which both install steps use).
- The ray conftest reclaims leaked placement groups between tests. It snapshots the table before each test and keeps what was already there, so a group a session- or module-scoped fixture legitimately holds isn't taken with them.
- The conftest's linux-only skip is overridable with `MLFORECAST_FORCE_RAY_TESTS=1`, so the suite can be run on macOS while developing.

The env-var workaround is removed from the workflow here, so CI now covers the supported configuration.

## ray on python 3.14

The ray extra was marker-gated to `python_version < '3.14'` and the conftest skipped the suite there. That gate is worth noting in this context: **ubuntu 3.14 was the only ubuntu job that stayed green while every other one hung — precisely because it was the one job not installing ray at all.** That green was masking the bug, not disproving it.

`ray` ships cp314 wheels as of 2.56.1, and `fugue[ray]`, `lightgbm` and `xgboost` all resolve on 3.14, so the gate is removed. The ubuntu 3.14 job now exercises the ray suite for the first time, which is a genuinely new path in CI.

## Tests

New `tests/distributed_ray/test_ray_models.py` covers:

- that LightGBM trains on the whole dataset across workers rather than on rank 0's shard — a constant feature and a split target, so the prediction is exactly the mean of whatever the model actually saw. It fails without the network params;
- that `model_` is the estimator the worker fitted: the user's `n_estimators` / `random_state` survive, `evals_result_` is populated, and its predictions match the booster's;
- that lightgbm's parameter aliases are honoured (`application`, `num_iterations`), which is what handing the translation back to the library buys;
- that xgboost keeps `random_state` and doesn't carry the ray callback back into `DistributedMLForecast.save`'s pickle;
- the `n_estimators=100` default, and that each worker gets a share of the cluster's CPUs;
- that a fit writes nothing to `~/ray_results` and honours an explicit `storage_path`;
- placement-group reclaim, in both directions: a leaked group is freed, a pre-existing one is left alone;
- that `clone()` preserves booster params — sklearn's introspection reads the subclass' signature, which only names the ray arguments, so `RayLGBMForecast._get_param_names` folds in `LGBMRegressor`'s. Without it the `clone()` in `_fit` would silently drop every parameter the user passed.

The skips for python < 3.10 are gone from this directory: `requires-python` is `>=3.10` and the CI matrix starts there, so none of them could fire.

### Runtime

These tests assert that predictions match across code paths, never model quality, so training 100 boosting rounds was pure cost — roughly 90% of the suite's runtime. They now pass `n_estimators=5`. The forecast test's data is also smaller: 20 series of 100-200 points with the lag transform at 20, rather than 100 series of 500-1000 with it at 335.

Together those take the ray suite from ~14m40s to under 2m, with every assertion unchanged. Note the split: the boosting rounds were nearly all of the win, and the data reduction only moved the heaviest test from ~28s to ~21s, because what remains is per-test ray orchestration rather than data volume. `--timeout=600` has a wide margin at these durations.

## Re-cutting the xgboost bound

`xgboost>=2.0.0,<2.1.0` existed for `xgboost_ray` compat, which is gone. The ray extra now takes `xgboost>=2.1.4`: `RayXGBForecast` mixes `RayForecastBase` into `XGBRegressor`, and `XGBModel.get_params` only survives walking into the mixin through the "if the immediate parent defines `get_params()`, use it; otherwise skip to `__bases__[1]`" fallback, which landed in 2.1.4. Below that it raises `AttributeError`, and `DistributedMLForecast._fit` clones the model, so it's a hard failure at fit time. It resolves to 3.4.1 on python >=3.12 and 3.2.0 below it.

Worth knowing this is **not** a ray-only change: the lock resolves one xgboost per python version, so that bound was constraining dask, spark and the local tests too. The APIs all three models rely on are unchanged across the major bump — the `save_raw("ubj")` -> `load_model` round trip and `xgboost.dask.DaskXGBRegressor`.

One cost to be aware of: xgboost 3.x depends on `nvidia-nccl` on linux (~250MB on 3.12+, ~340MB below), so the ubuntu jobs download that once per lock change. macOS and Windows are unaffected — the dependency is marked `sys_platform == 'linux'`.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -94,8 +94,5 @@ jobs:
       - name: Run pytest
         env:
           RAY_ENABLE_UV_RUN_RUNTIME_ENV: 0
-          # ray >=2.56 converts blocks to pandas with pd.ArrowDtype columns, which
-          # lightgbm rejects ("pandas dtypes must be int, float or bool"). Keep the
-          # pre-2.56 numpy dtypes until the ray models cast the features themselves.
-          RAY_DATA_ENABLE_ARROW_BACKED_PANDAS_CONVERSION: 0
-        run: uv run pytest --reruns=2 --reruns-delay=2 --only-rerun="Socket recv error"
+        # --timeout keeps a hung ray test from burning the job's full 6h budget
+        run: uv run pytest --timeout=600 --reruns=2 --reruns-delay=2 --only-rerun="Socket recv error"
--- a/docs/distributed.models.ray.lgb.html.md
+++ b/docs/distributed.models.ray.lgb.html.md
@@ -5,8 +5,8 @@ title: RayLGBMForecast
 ---
 
 
-Wrapper of `lightgbm.ray.RayLGBMRegressor` that adds a `model_` property
-that contains the fitted booster and is sent to the workers to in the
-forecasting step.
+LightGBM forecaster trained with `ray.train.lightgbm.LightGBMTrainer`. Adds a
+`model_` property that contains the fitted booster as a local
+`lightgbm.LGBMRegressor` and is sent to the workers in the forecasting step.
 
 ::: mlforecast.distributed.models.ray.lgb.RayLGBMForecast
--- a/docs/distributed.models.ray.lgb.html.md
+++ b/docs/distributed.models.ray.lgb.html.md
@@ -6,7 +6,7 @@ title: RayLGBMForecast
 
 
 LightGBM forecaster trained with `ray.train.lightgbm.LightGBMTrainer`. Adds a
-`model_` property that contains the fitted booster as a local
+`model_` attribute that contains the fitted booster as a local
 `lightgbm.LGBMRegressor` and is sent to the workers in the forecasting step.
 
 ::: mlforecast.distributed.models.ray.lgb.RayLGBMForecast
--- a/docs/distributed.models.ray.xgb.html.md
+++ b/docs/distributed.models.ray.xgb.html.md
@@ -5,8 +5,8 @@ title: RayXGBForecast
 ---
 
 
-Wrapper of `xgboost.ray.RayXGBRegressor` that adds a `model_` property
-that contains the fitted model and is sent to the workers in the
-forecasting step.
+XGBoost forecaster trained with `ray.train.xgboost.XGBoostTrainer`. Adds a
+`model_` property that contains the fitted booster as a local
+`xgboost.XGBRegressor` and is sent to the workers in the forecasting step.
 
 ::: mlforecast.distributed.models.ray.xgb.RayXGBForecast
--- a/docs/distributed.models.ray.xgb.html.md
+++ b/docs/distributed.models.ray.xgb.html.md
@@ -6,7 +6,7 @@ title: RayXGBForecast
 
 
 XGBoost forecaster trained with `ray.train.xgboost.XGBoostTrainer`. Adds a
-`model_` property that contains the fitted booster as a local
+`model_` attribute that contains the fitted booster as a local
 `xgboost.XGBRegressor` and is sent to the workers in the forecasting step.
 
 ::: mlforecast.distributed.models.ray.xgb.RayXGBForecast
--- a/mlforecast/distributed/forecast.py
+++ b/mlforecast/distributed/forecast.py
@@ -29,7 +29,6 @@ try:
 except ModuleNotFoundError:
     SPARK_INSTALLED = False
 try:
-    from lightgbm_ray import RayDMatrix
     from ray.data import Dataset as RayDataset
 
     RAY_INSTALLED = True
@@ -435,12 +434,8 @@ class DistributedMLForecast:
             prep_selected = prep.select_columns(
                 cols=features + [target_col]
             ).materialize()
-            X = RayDMatrix(
-                prep_selected,
-                label=target_col,
-            )
             for name, model in self.models.items():
-                trained_model = clone(model).fit(X, y=None)
+                trained_model = clone(model).fit(prep_selected, target_col=target_col)
                 self.models_[name] = trained_model.model_
         else:
             raise NotImplementedError(
--- a/mlforecast/distributed/forecast.py
+++ b/mlforecast/distributed/forecast.py
@@ -426,7 +426,8 @@ class DistributedMLForecast:
                 trained_model = clone(model).fit(X, y, sample_weight=weights)
                 self.models_[name] = trained_model.model_
         elif RAY_INSTALLED and isinstance(data, RayDataset):
-            # Need to materialize
+            # Need to materialize. Each model's fit would otherwise re-execute
+            # this same dataset, since they all get handed the lazy one.
             if weight_col is not None:
                 raise NotImplementedError(
                     "Only spark and dask engines currently support sample weights."
--- a/mlforecast/distributed/models/ray/_base.py
+++ b/mlforecast/distributed/models/ray/_base.py
@@ -1,0 +1,87 @@
+__all__ = ["RayForecastBase"]
+
+
+from typing import Any, Callable, Dict, Optional, Tuple
+
+from sklearn.base import BaseEstimator
+
+
+class RayForecastBase(BaseEstimator):
+    """Base for the ray models, trained through ray.train's Trainer API.
+
+    Subclasses provide ``_trainer_cls``, ``_train_loop`` (a module level function,
+    so that it can be shipped to the workers), ``_translate_params`` and
+    ``_local_model``.
+
+    The ray specific arguments are keyword only so that they can't collide with
+    the booster's parameters, which are taken as ``**params``.
+    """
+
+    _trainer_cls: Any
+    _train_loop: Callable[[Dict[str, Any]], None]
+
+    def __init__(
+        self,
+        *,
+        num_workers: int = 1,
+        resources_per_worker: Optional[Dict[str, float]] = None,
+        **params: Any,
+    ):
+        self.num_workers = num_workers
+        self.resources_per_worker = resources_per_worker
+        self.params = params
+
+    # sklearn's introspection ignores **kwargs, which would make `clone` drop
+    # every booster parameter, so the params are handled explicitly here.
+    def get_params(self, deep: bool = True) -> Dict[str, Any]:  # noqa: ARG002 - sklearn's signature
+        return {
+            "num_workers": self.num_workers,
+            "resources_per_worker": self.resources_per_worker,
+            **self.params,
+        }
+
+    def set_params(self, **params: Any) -> "RayForecastBase":
+        for name in ("num_workers", "resources_per_worker"):
+            if name in params:
+                setattr(self, name, params.pop(name))
+        self.params.update(params)
+        return self
+
+    def _translate_params(self) -> Tuple[Dict[str, Any], int]:
+        """Return the booster params and the number of boosting rounds."""
+        raise NotImplementedError
+
+    def _local_model(self, checkpoint: Any, params: Dict[str, Any]) -> Any:
+        """Rebuild a local (non-distributed) model from the training checkpoint."""
+        raise NotImplementedError
+
+    def fit(self, dataset: Any, target_col: str) -> "RayForecastBase":
+        from ray.train import ScalingConfig
+
+        params, num_boost_round = self._translate_params()
+        trainer = self._trainer_cls(
+            self._train_loop,
+            train_loop_config={
+                "params": params,
+                "num_boost_round": num_boost_round,
+                "target_col": target_col,
+            },
+            scaling_config=ScalingConfig(
+                num_workers=self.num_workers,
+                resources_per_worker=self.resources_per_worker,
+            ),
+            datasets={"train": dataset},
+        )
+        result = trainer.fit()
+        self.model_ = self._local_model(result.checkpoint, params)
+        return self
+
+    def _pop_num_boost_round(self, params: Dict[str, Any], *aliases: str) -> int:
+        for alias in aliases:
+            if alias in params:
+                num_boost_round = params.pop(alias)
+                # drop any remaining aliases so they don't reach the booster
+                for other in aliases:
+                    params.pop(other, None)
+                return int(num_boost_round)
+        return 100
--- a/mlforecast/distributed/models/ray/_base.py
+++ b/mlforecast/distributed/models/ray/_base.py
@@ -1,87 +1,116 @@
 __all__ = ["RayForecastBase"]
 
 
-from typing import Any, Callable, Dict, Optional, Tuple
+import pickle
+import tempfile
+from pathlib import Path
+from typing import Any, Callable, Dict, Optional
 
-from sklearn.base import BaseEstimator
+_RAY_PARAMS = ("num_workers", "resources_per_worker")
+_MODEL_FILE = "model.pkl"
 
 
-class RayForecastBase(BaseEstimator):
-    """Base for the ray models, trained through ray.train's Trainer API.
+def worker_n_jobs(requested: Any) -> int:
+    """Threads for the booster, from the CPUs this train worker was actually given.
 
-    Subclasses provide ``_trainer_cls``, ``_train_loop`` (a module level function,
-    so that it can be shipped to the workers), ``_translate_params`` and
-    ``_local_model``.
+    Both ``lightgbm_ray`` and ``xgboost_ray`` sized the thread pool from the
+    actor's CPU share; without it N workers landing on one node each spawn
+    threads for every core on the box. An explicit, smaller value is honoured.
+    """
+    import ray
+
+    assigned = ray.get_runtime_context().get_assigned_resources().get("CPU", 1)
+    assigned = max(1, int(assigned))
+    try:
+        requested = int(requested)
+    except (TypeError, ValueError):
+        return assigned
+    return assigned if requested <= 0 else min(requested, assigned)
+
+
+class KeepLastMetrics:
+    """Remember the last iteration's metrics so the final report can carry them."""
+
+    last_metrics: Dict[str, Any] = {}
+
+    def _report_metrics(self, report_dict: Dict[str, Any]) -> None:
+        self.last_metrics = report_dict
+        super()._report_metrics(report_dict)  # type: ignore[misc]
+
+
+def report_fitted_model(
+    model: Any, booster: Any, booster_file: str, metrics: Dict[str, Any]
+) -> None:
+    """Report ray's standard booster artifact along with the fitted estimator.
+
+    The estimator is what becomes ``model_``, which is why it's checkpointed;
+    the booster is kept next to it so that ``RayTrainReportCallback.get_model``
+    still works on the result.
+
+    ``ray.train.report`` is collective, so every worker has to call it;
+    reporting from rank 0 only deadlocks.
+    """
+    import ray.train
+    from ray.train import Checkpoint
+
+    if ray.train.get_context().get_world_rank() != 0:
+        ray.train.report(metrics)
+        return
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        booster.save_model(Path(tmp_dir, booster_file).as_posix())
+        with open(Path(tmp_dir, _MODEL_FILE), "wb") as f:
+            pickle.dump(model, f)
+        ray.train.report(metrics, checkpoint=Checkpoint.from_directory(tmp_dir))
+
+
+class RayForecastBase:
+    """Mixin holding the ray.train plumbing; subclasses are real sklearn estimators.
+
+    The training loop builds and fits the library's own estimator in the worker,
+    as ``lightgbm.dask._train_part`` does, and checkpoints it. That keeps all of
+    the parameter handling in the library instead of here.
 
     The ray specific arguments are keyword only so that they can't collide with
-    the booster's parameters, which are taken as ``**params``.
+    the booster's parameters, which are taken as ``**kwargs``.
     """
 
-    _trainer_cls: Any
-    _train_loop: Callable[[Dict[str, Any]], None]
+    num_workers: int
+    resources_per_worker: Optional[Dict[str, float]]
 
     def __init__(
         self,
         *,
         num_workers: int = 1,
         resources_per_worker: Optional[Dict[str, float]] = None,
-        **params: Any,
+        **kwargs: Any,
     ):
+        # cooperative: goes on to LGBMRegressor / XGBRegressor
+        super().__init__(**kwargs)
         self.num_workers = num_workers
         self.resources_per_worker = resources_per_worker
-        self.params = params
 
-    # sklearn's introspection ignores **kwargs, which would make `clone` drop
-    # every booster parameter, so the params are handled explicitly here.
-    def get_params(self, deep: bool = True) -> Dict[str, Any]:  # noqa: ARG002 - sklearn's signature
-        return {
-            "num_workers": self.num_workers,
-            "resources_per_worker": self.resources_per_worker,
-            **self.params,
-        }
-
-    def set_params(self, **params: Any) -> "RayForecastBase":
-        for name in ("num_workers", "resources_per_worker"):
-            if name in params:
-                setattr(self, name, params.pop(name))
-        self.params.update(params)
-        return self
-
-    def _translate_params(self) -> Tuple[Dict[str, Any], int]:
-        """Return the booster params and the number of boosting rounds."""
-        raise NotImplementedError
-
-    def _local_model(self, checkpoint: Any, params: Dict[str, Any]) -> Any:
-        """Rebuild a local (non-distributed) model from the training checkpoint."""
-        raise NotImplementedError
-
-    def fit(self, dataset: Any, target_col: str) -> "RayForecastBase":
+    def _train(
+        self,
+        trainer_cls: Any,
+        train_loop: Callable[[Dict[str, Any]], None],
+        dataset: Any,
+        target_col: str,
+    ) -> "RayForecastBase":
         from ray.train import ScalingConfig
 
-        params, num_boost_round = self._translate_params()
-        trainer = self._trainer_cls(
-            self._train_loop,
-            train_loop_config={
-                "params": params,
-                "num_boost_round": num_boost_round,
-                "target_col": target_col,
-            },
+        params = self.get_params()  # type: ignore[attr-defined]
+        for name in _RAY_PARAMS:
+            params.pop(name, None)
+        trainer = trainer_cls(
+            train_loop,
+            train_loop_config={"params": params, "target_col": target_col},
             scaling_config=ScalingConfig(
                 num_workers=self.num_workers,
                 resources_per_worker=self.resources_per_worker,
             ),
             datasets={"train": dataset},
         )
-        result = trainer.fit()
-        self.model_ = self._local_model(result.checkpoint, params)
+        with trainer.fit().checkpoint.as_directory() as ckpt_dir:
+            with open(Path(ckpt_dir, _MODEL_FILE), "rb") as f:
+                self.model_ = pickle.load(f)
         return self
-
-    def _pop_num_boost_round(self, params: Dict[str, Any], *aliases: str) -> int:
-        for alias in aliases:
-            if alias in params:
-                num_boost_round = params.pop(alias)
-                # drop any remaining aliases so they don't reach the booster
-                for other in aliases:
-                    params.pop(other, None)
-                return int(num_boost_round)
-        return 100
--- a/mlforecast/distributed/models/ray/_base.py
+++ b/mlforecast/distributed/models/ray/_base.py
@@ -99,14 +99,32 @@ class RayForecastBase:
         ``ScalingConfig`` assigns a single CPU per worker when this isn't set,
         which would make a default fit single threaded. ``xgboost_ray`` split the
         cluster's CPUs across its actors instead (``_autodetect_resources``);
-        that's kept here so that the default isn't a slowdown.
+        that's kept here so that the default isn't a slowdown, bound by the
+        smallest node included.
+
+        The share also becomes ray data's ``exclude_resources``
+        (``DataParallelTrainer`` hands it ``scaling_config.total_resources``), so
+        a worker that takes every CPU on its node leaves data none to execute
+        with. ``_train`` materializes before building the trainer so that there
+        is nothing left for data to do by then.
         """
         import ray
 
         if self.resources_per_worker is not None:
             return self.resources_per_worker
         cpus = int(ray.cluster_resources().get("CPU", 1))
-        return {"CPU": max(1, cpus // self.num_workers)}
+        # a placement group bundle has to fit on a single node, so the cluster
+        # wide share is bounded by the smallest one as well
+        min_node_cpus = min(
+            (
+                node.get("Resources", {}).get("CPU", 0.0)
+                for node in ray.nodes()
+                if node.get("Alive", False)
+            ),
+            default=0.0,
+        )
+        share = min(int(min_node_cpus or 1), cpus // self.num_workers)
+        return {"CPU": max(1, share)}
 
     def _train(
         self,
@@ -120,6 +138,13 @@ class RayForecastBase:
         params = self.get_params()  # type: ignore[attr-defined]
         for name in _RAY_PARAMS:
             params.pop(name, None)
+        # execute the dataset before the trainer exists. ray train reserves the
+        # workers' CPUs away from ray data (`ScalingConfig.total_resources`
+        # becomes data's `exclude_resources`), so a dataset with work still
+        # pending once the placement group holds them has nothing left to run
+        # with and blocks forever. Nothing is given up by doing it here: the
+        # train loops build a `Dataset`/`DMatrix` from the whole shard anyway.
+        dataset = dataset.materialize()
         with contextlib.ExitStack() as stack:
             storage_path = self.storage_path
             if storage_path is None:
--- a/mlforecast/distributed/models/ray/_base.py
+++ b/mlforecast/distributed/models/ray/_base.py
@@ -1,12 +1,13 @@
 __all__ = ["RayForecastBase"]
 
 
+import contextlib
 import pickle
 import tempfile
 from pathlib import Path
 from typing import Any, Callable, Dict, Optional
 
-_RAY_PARAMS = ("num_workers", "resources_per_worker")
+_RAY_PARAMS = ("num_workers", "resources_per_worker", "storage_path")
 _MODEL_FILE = "model.pkl"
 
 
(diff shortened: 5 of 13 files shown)


jmoralez:
Thanks for the thorough writeup — the diagnosis is right, `ray.train` is the correct target, and the placement-group analysis explains the CI behaviour well. I went through it locally and there's one blocking correctness bug plus a few smaller ones, and a structural suggestion that would make most of `_base.py` unnecessary.

Everything below was verified on ray 2.56.1 / lightgbm 4.7.0 / xgboost 3.4.1 and 2.0.3, python 3.12.

---

# Bugs

## 1. LightGBM silently trains on 1/N of the data when `num_workers > 1` — blocking

`_lgb_train_loop` calls `lgb.train` without lightgbm's network parameters. Ray's `_LightGBMBackend` doesn't inject them: `on_training_start` only stashes them in a per-worker global that the train loop has to pull in itself, which is why ray's own `LightGBMTrainer` docstring example has this in it:

```python
params = {
    "objective": "regression",
    # Adding the lines below are the only changes needed
    # for your `lgb.train` call!
    "tree_learner": "data_parallel",
    "pre_partition": True,
    **ray.train.lightgbm.get_network_params(),
}
```

Without them, each worker trains an independent local model on its own shard and `RayTrainReportCallback` checkpoints rank 0's — so `model_` is a model fit on `1/num_workers` of the data, with no error anywhere.

Repro — constant feature, `y=0` on the first half and `y=100` on the second, so with no possible split the prediction is exactly the mean of whatever the model actually saw:

```python
df = pd.DataFrame({"x": np.ones(200), "y": np.r_[np.zeros(100), np.full(100, 100.0)]})
# ... LightGBMTrainer(_lgb_train_loop, scaling_config=ScalingConfig(num_workers=2), ...)
```

```
this PR's train loop                          ->  0.0     # rank 0's shard only
same loop + get_network_params()              -> 50.0     # correct
```

XGBoost is unaffected — `XGBoostConfig.train_func_context` wraps the whole train loop in a `CommunicatorContext`, so `xgb.train` is distributed automatically. This is a lightgbm-only fix.

**Also needs a test.** `num_workers=2` currently appears only in `test_clone_preserves_booster_params`, which never trains; every test that actually fits uses the default `num_workers=1`, so nothing in the suite would have caught this. The constant-feature setup above is cheap and deterministic and would pin it.

## 2. `params.setdefault("objective", "regression")` silently discards lightgbm's objective aliases

`objective` has five aliases in lightgbm (`app`, `application`, `loss`, `objective`, `objective_type`). `setdefault` only checks the canonical name, and `lgb.train`'s `_choose_param_value` prefers the canonical name when both are present, so the user's objective is dropped:

```python
>>> RayLGBMForecast(application="poisson", verbosity=-1)._translate_params()[0]
{'application': 'poisson', 'verbosity': -1, 'objective': 'regression'}
>>> # objective lightgbm actually trains with, given those params: 'regression'
>>> # ... and without the setdefault:                              'poisson'
```

## 3. The `random_state` -> `seed` translation for xgboost is unnecessary

The comment says "the native API doesn't know about the sklearn parameter names", but it does. On both 2.0.3 and 3.4.1, `xgb.train({"random_state": 1})` emits no warning and produces a bit-identical booster to `{"seed": 1}`, and a different one for a different value. xgboost's own `XGBModel.get_xgb_params()` passes `random_state` straight through to `train()` — that's the code path every `XGBRegressor.fit` call already takes.

## 4. `_pop_num_boost_round` covers 3 of lightgbm's 11 aliases, and its precedence differs from lightgbm's

`_ConfigAliases.get("num_iterations")` is `{max_iter, n_estimators, n_iter, nrounds, num_boost_round, num_iteration, num_iterations, num_round, num_rounds, num_tree, num_trees}`. The eight not listed fall through into `params` — harmless for lightgbm, because `lgb.train` resolves them itself, which is also what makes the helper redundant on that side. Where it does bite: when two aliases are passed, this picks the first in *its* list, whereas `_choose_num_iterations` prefers the canonical `num_iterations` and warns about the conflict.

## 5. `model_` loses the user's params and the booster's scores

`_translate_params` pops `n_estimators` before `_local_model` builds the estimator, so the returned model reports the default:

```
asked n_estimators=5 -> model_.n_estimators: 100 | booster trees: 5 | model_.best_score_: {}
```

Predictions are correct (the booster is grafted), but introspection on `model_` isn't. `_best_score`/`_evals_result` are blanked rather than read off the booster. Separately, 6 of the 11 assignments in the graft re-set what `LGBMRegressor.__init__` has already set, and `_le` / `_class_map` / `_classes` / `_n_classes` are classifier-only state on a regressor.

## 6. Thread oversubscription on multi-worker nodes

`xgboost_ray` set `nthread` per actor from `cpus_per_actor` (`main.py:679-685`) and `lightgbm_ray` did the equivalent via `_set_omp_num_threads`. Nothing sets it now, so N workers landing on one node each spawn threads for every core on the box.

## 7. The compatibility note needs one more line

"Constructors are unchanged" holds for booster params, but the worker count moved: both old wrappers derived it from `n_jobs` via `RayParams(num_actors=n_jobs)` (falling back to 1). The default is unchanged, but anyone who set `n_jobs=8` to get 8 actors now silently gets one worker while `n_jobs` becomes a thread count. That belongs in the compatibility section.

---

# Suggestion: fit the real estimator in the worker, as `lightgbm.dask` does

The custom `get_params`/`set_params`, the alias tables and the attribute graft all trace back to one decision: `RayForecastBase` derives from a bare `BaseEstimator`, and the checkpoint carries a bare `Booster`. On `main` neither was true — `RayLGBMForecast` was ultimately an `LGBMRegressor`, which is *why* `clone()` worked and why there was no param-translation code to write.

Both libraries solve both halves already, and this is exactly how they write their own distributed estimators. `lightgbm.dask` never synthesizes fitted state: `_train_part` builds a real estimator in the worker and returns it fitted (`lightgbm/dask.py:321-360`)

```python
model = model_factory(**params)   # params carries machines/local_listen_port/num_machines/tree_learner
model.fit(data, label, ...)
return model if return_model else None
```

and `_lgb_dask_fit` then copies from *that fitted estimator*. The ray equivalent is to put it in the checkpoint. Nothing has to be given up to do it — the checkpoint can carry ray's standard artifact **and** the estimator, so `RayTrainReportCallback` keeps reporting per-iteration metrics and `RayTrainReportCallback.get_model` still works on the result.

`_base.py` becomes the trainer wiring plus one helper:

```python
class KeepLastMetrics:
    """Remember the last iteration's metrics so the final report can carry them."""

    last_metrics: Dict[str, Any] = {}

    def _report_metrics(self, report_dict: Dict[str, Any]) -> None:
        self.last_metrics = report_dict
        super()._report_metrics(report_dict)


def report_fitted_model(model, booster, booster_file: str, metrics: Dict[str, Any]) -> None:
    """Report ray's standard booster artifact along with the fitted estimator.

    ``ray.train.report`` is collective, so every worker has to call it.
    """
    if ray.train.get_context().get_world_rank() != 0:
        ray.train.report(metrics)
        return
    with tempfile.TemporaryDirectory() as tmp_dir:
        booster.save_model(Path(tmp_dir, booster_file).as_posix())
        with open(Path(tmp_dir, _MODEL_FILE), "wb") as f:
            pickle.dump(model, f)
        ray.train.report(metrics, checkpoint=Checkpoint.from_directory(tmp_dir))


class RayForecastBase:
    """Mixin holding the ray.train plumbing; subclasses are real sklearn estimators."""

    def __init__(self, *, num_workers=1, resources_per_worker=None, **kwargs):
        super().__init__(**kwargs)
        self.num_workers = num_workers
        self.resources_per_worker = resources_per_worker

    def _train(self, trainer_cls, train_loop, dataset, target_col):
        params = self.get_params()
        for name in _RAY_PARAMS:
            params.pop(name)
        trainer = trainer_cls(
            train_loop,
            train_loop_config={"params": params, "target_col": target_col},
            scaling_config=ScalingConfig(num_workers=self.num_workers,
                                         resources_per_worker=self.resources_per_worker),
            datasets={"train": dataset},
        )
        with trainer.fit().checkpoint.as_directory() as ckpt_dir:
            with open(Path(ckpt_dir, _MODEL_FILE), "rb") as f:
                self.model_ = pickle.load(f)
        return self
```

and `lgb.py` is the whole model:

```python
class _ReportCallback(KeepLastMetrics, RayTrainReportCallback):
    pass


def _lgb_train_loop(config):
    df = normalize_pandas_for_lightgbm(ray.train.get_dataset_shard("train").materialize().to_pandas())
    label = df.pop(config["target_col"])
    callback = _ReportCallback(checkpoint_at_end=False)
    # network params are plain kwargs, as in lightgbm.dask's _train_part
    model = lgb.LGBMRegressor(**config["params"], tree_learner="data_parallel", **get_network_params())
    model.fit(df, label, eval_set=[(df, label)], eval_names=["train"], callbacks=[callback])
    report_fitted_model(model, model.booster_, _ReportCallback.CHECKPOINT_NAME, callback.last_metrics)


class RayLGBMForecast(RayForecastBase, lgb.LGBMRegressor):
    @classmethod
    def _get_param_names(cls):
        # sklearn reads the subclass' signature, which doesn't name the booster params
        return sorted([*lgb.LGBMRegressor._get_param_names(), *_RAY_PARAMS])

    def fit(self, dataset, target_col):
        return self._train(LightGBMTrainer, _lgb_train_loop, dataset, target_col)
```

`xgb.py` is the same shape with `get_booster()` instead of `booster_` and no `_get_param_names` override, since `XGBModel.get_params` already walks base classes.

Verified end to end with `num_workers=2`:

```
RayLGBMForecast  pred(want~50)= 50.00 type=LGBMRegressor n_estimators=5 random_state=0 eval_metrics_kept=True
RayXGBForecast   pred(want~50)= 50.00 type=XGBRegressor  n_estimators=5 random_state=0 eval_metrics_kept=True
checkpoint contents: ['model.pkl', 'model.txt']
RayTrainReportCallback.get_model still works -> Booster | trees: 1 | pred: 50.0
per-iteration metrics reported: {'train-l2': 2500.0}
```

**Pros**

- Bugs 2-5 stop existing rather than getting patched. `get_params`/`set_params`, `_translate_params`, `_pop_num_boost_round` and the graft all go: the worker's own `.fit()` does the param translation, and the estimator it returns *is* `model_`.
- Zero private-attribute writes. The only private API left is the one-line `_get_param_names` override on the lightgbm side; `LGBMModel.get_params` / `XGBModel.get_params` / `get_xgb_params` are the libraries' own supported surface.
- `model_` improves: it keeps the user's params (`n_estimators=5`, not 100) and its `evals_result_`, instead of being rebuilt from native params.
- Keeps `isinstance(model, LGBMRegressor)` / `XGBRegressor` true, as on `main`, so the "no longer a subclass" break in the compatibility section goes away.
- Smaller: 127 lines across the three files vs. 222 here.

**Cons**

- The checkpoint gains a pickle next to `model.txt`. It's interpreter- and version-sensitive if anyone ever persists and reloads it out of band — `model.txt` is still there for that, but it is an extra artifact with different durability guarantees.
- Two reports per run (the callback's per-iteration metrics, then the final one). `ray.train.report` is collective, so every rank has to call it — that's what the early return in `report_fitted_model` is for. A rank-0-only report deadlocks.
- Cooperative multiple inheritance: `RayForecastBase.__init__` has to `super().__init__(**kwargs)` and the MRO order matters.
- `fit(dataset, target_col)` still doesn't match the sklearn `fit(X, y)` contract it nominally inherits, and `predict`/`score` are inherited but unusable pre-`fit`. Both were equally true on `main` (`RayLGBMRegressor.fit(X, y=None)` with a `RayDMatrix`), so not a regression, but worth a docstring line.

Two gotchas that each cost me a run, worth knowing before writing it:

- `XGBModel.get_params()` already contains a `callbacks` key, so `xgb.XGBRegressor(**params, callbacks=[cb])` raises `got multiple values for keyword argument 'callbacks'`. Merge it into the dict instead.
- xgboost *stores* `callbacks` as a param, so the fitted estimator carries the ray callback back to the driver and into whatever the user pickles via `DistributedMLForecast.save`. Call `model.set_params(callbacks=None)` before reporting. LightGBM takes callbacks in `fit` and doesn't retain them, so it's xgboost-only.

**Middle ground, if the full rework is too much for this PR:** keep the current structure and just swap the hand-rolled translation for the libraries' helpers — `get_xgb_params()` / `get_num_boosting_rounds()` on the xgboost side, `_choose_param_value("num_iterations", params, 100)` on the lightgbm side. *Pro:* small diff, fixes bugs 2-4. *Con:* the custom `get_params`/`set_params` and the graft stay, so bug 5 and the rot risk stay with them.

---

# Smaller suggestions

**`_reclaim_placement_groups` removes every non-`REMOVED` group in the cluster, not just leaked ones.** *Pro:* dead simple and correct for the suite as it stands. *Con:* it will silently break the first fixture that legitimately holds a placement group across tests. Filtering to groups created by the trainer, or scoping it to a session-level teardown, would be sturdier.

**`_trainer_cls` is annotated as a class attribute on the base but implemented as a `@property` in both subclasses.** *Pro:* gives lazy importing. *Con:* the annotation is misleading and a property that ignores `self` is an odd shape. A plain import inside `fit` reads better.

**`trainer.fit()` writes to `~/ray_results` per model per call under the default `RunConfig`.** Neither old wrapper did. Worth deciding whether to point `storage_path` somewhere or document it, since `DistributedMLForecast.fit` calls it once per model.

**Scope.** Four independent changes are riding together here. The one I'd most like split out is dropping `xgboost<2.1.0`: as the description notes, the lock resolves one xgboost per python version, so that bound was constraining dask, spark and the local tests too — it's a repo-wide bump landing inside a ray PR, and it deserves its own bisect point. The py3.14 enablement is likewise independently reviewable.

jmoralez:
Reviewed e39dc3f. The rework is what I was hoping for — the estimator comes back from the worker, so the alias handling, the graft and the custom `get_params` are all gone rather than patched, and `test_lgb_trains_on_the_full_dataset_across_workers` genuinely fails without the network params. Ran the suite locally on ray 2.56.1 / lightgbm 4.7.0 / xgboost 3.4.1, python 3.12: **14 passed** (`test_ray_models.py` alone: 9 passed).

Two things to fix before merge.

## 1. The ray extra needs `xgboost>=2.1.4`

`RayXGBForecast(RayForecastBase, xgb.XGBRegressor)` depends on `XGBModel.get_params`'s "if the immediate parent defines `get_params()`, use it; otherwise skip to `__bases__[1]`" fallback. `RayForecastBase` is a plain mixin with no `get_params`, so older xgboost takes `__bases__[0]` unconditionally:

```
xgboost 2.0.3
get_params FAILED: AttributeError type object 'RayForecastBase' has no attribute 'get_params'
```

That fallback landed in xgboost 2.1.4 (absent in 2.1.3). `DistributedMLForecast._fit` calls `clone(model)`, so this is a hard failure at fit time. The extra is now plain `"xgboost; sys_platform != 'win32'"` with no floor, and the lock resolves 3.4.1/3.2.0, so CI won't catch it. Slightly ironic given this PR drops `>=2.0.0,<2.1.0` — the new code needs a *higher* floor than the one that was removed.

## 2. Default training is single-threaded, and `n_jobs` can no longer raise it

`worker_n_jobs` sizes the pool from the worker's assigned CPUs, and `ScalingConfig` assigns `{"CPU": 1}` when `resources_per_worker` is None (`ray/air/config.py`, `_resources_per_worker_not_none`). Measured with `ray.init(num_cpus=8)`:

```
{}                                                -> model_.n_jobs = 1
{'n_jobs': 8}                                     -> model_.n_jobs = 1
{'resources_per_worker': {'CPU': 4}}              -> model_.n_jobs = 4
{'n_jobs': 2, 'resources_per_worker': {'CPU': 4}} -> model_.n_jobs = 2
```

The mechanism is right and the escape hatch works. The problem is the baseline it replaces: `xgboost_ray._autodetect_resources` gave the default single actor `cluster_cpus // num_actors`, i.e. every CPU in the cluster. So `RayLGBMForecast()` goes from all cores to one thread, and the obvious remedy (`n_jobs=8`) silently does nothing.

The compatibility note says "an `n_jobs` larger than that share is capped", which describes row 2 but undersells it — `n_jobs` alone can now never *raise* the thread count. Either default `resources_per_worker` by mirroring `_autodetect_resources`, or say plainly in the docstring that `resources_per_worker` is the CPU knob and `n_jobs` only caps below it.

**Nit on the same code:** the clamp persists into `model_`, which is shipped for prediction and returned by `to_local()`, so a user's `n_jobs=8` is 1 there too. Restoring the user's value on the estimator before reporting would fit the same slot where `callbacks` is cleared for xgboost.

## Still open from the previous round, all fine to defer

- `_reclaim_placement_groups` removes every group in the cluster, not just leaked ones — it'll break the first fixture that legitimately holds one.
- No `storage_path`, so `~/ray_results` still grows per model per fit.
- `docs/distributed.models.ray.{lgb,xgb}.html.md` still describe `model_` as a "property"; it's an attribute now.
- The xgboost-bound and py3.14 changes are still bundled in with the fix.

jmoralez:
Reviewed `4f35d2c`. Everything from the last round landed: the `xgboost>=2.1.4` floor is in pyproject and the lock, `resources_per_worker` now sizes the CPUs, the clamp no longer persists into `model_` (checked both models — `n_jobs=8` comes back as 8), and the three deferred items — reclaim scoping, `storage_path`, the docs "property" wording — are all done. Ran the suite on ray 2.56.1 / lightgbm 4.6.0 / xgboost 3.4.1, python 3.14: **17 passed**. I also confirmed `worker_n_jobs` behaves as documented inside a real train worker, and that `ScalingConfig(resources_per_worker=None)` really does hand out `{"CPU": 1}`.

One new blocker, in the fix itself.

## `_resources_per_worker` hangs `fit()` on a multi-node cluster

It sizes the bundle from `ray.cluster_resources()`, but a placement group bundle has to fit on a **single** node. At the default `num_workers=1` that asks for every CPU in the cluster:

```
cluster_resources CPU: 4.0        # two nodes, 2 CPUs each
requested resources_per_worker: {'CPU': 4}
(autoscaler) Error: No available node types can fulfill resource request {'CPU': 4.0}
HUNG (no return within 120s)
```

Same at `num_workers>1` on heterogeneous nodes (1 and 7 CPUs, 2 workers → 4 per bundle, unschedulable). This is the failure mode the PR set out to remove, and `--timeout=600` doesn't reach users.

`_autodetect_resources` has the missing half — it bounds the share by the smallest node (`main.py:837-845`):

```python
cpus_per_actor = max(1, min(int(_get_min_node_cpus() or 1),
                            int(cluster_cpus // ray_params.num_actors)))
```

With that `min` the same cluster requests `{'CPU': 2}` and the fit completes. The docstring's "that's kept here" should follow.

**Two nits on the same commit.** `test_lgb_trains_on_the_full_dataset_across_workers:72` used to assert `model_.n_jobs == 1`, which exercised the clamp; the restore made it `== 2`, so `worker_n_jobs` now has no coverage and could silently regress to a constant. And `test_workers_get_a_share_of_the_cluster_cpus:200` pins the unbounded value, so it needs updating with the fix.

nasaul:
Fixed in da9593b. The blocker is gone, and investigating it turned up a second hang in the same function that the requested `min` doesn't cover.

## 1. The bundle has to fit on one node

Restored the half of `_autodetect_resources` we dropped:

```python
min_node_cpus = min(
    (n.get("Resources", {}).get("CPU", 0.0) for n in ray.nodes() if n.get("Alive", False)),
    default=0.0,
)
share = min(int(min_node_cpus or 1), cpus // self.num_workers)
```

Verified on a real two node cluster (`ray.cluster_utils.Cluster`, 2 CPUs each), default `num_workers=1` — the exact configuration that hung on `4f35d2c`:

```
cluster CPU: 4.0
alive nodes: [2.0, 2.0]
requested: {'CPU': 2}      # was {'CPU': 4} -> unschedulable -> hung
OK pred: [50.]
```

## 2. Sizing from the cluster also starves ray data — and this one is single node

`DataParallelTrainer` hands `ScalingConfig.total_resources` to ray data as `exclude_resources` (`data_parallel_trainer.py:262-266` → `_internal/data_config.py:126-135`). At `num_workers=1` the default share is every CPU on the node, so data's budget is `cluster − cluster = 0` and any dataset with work still pending blocks forever:

```
MapBatches(<lambda>): 0/1
  Tasks: 1 [backpressured:tasks(ResourceBudget)]
```

Measured on one 4-CPU node:

| dataset | bundle | result |
|---|---|---|
| lazy (`map_batches`) | `{"CPU": 4}` | **hung** |
| lazy | `{"CPU": 1}` | ok |
| materialized | `{"CPU": 4}` | ok |

`min_node_cpus == cluster_cpus` on a single node, so the fix above does nothing for it. It isn't reachable through `DistributedMLForecast` — `forecast.py` materializes before fitting — but only that one call three files away was holding it off, and `RayLGBMForecast` is public.

**Fix:** `_train` materializes before the trainer is built, while nothing holds a placement group.

```python
dataset = dataset.materialize()
```

### Why that and not a CPU reservation

Reserving a CPU per worker for data was the obvious alternative and it's worse: it taxes every fit, including the `DistributedMLForecast` path that can't hit the deadlock, and "is one CPU enough for data?" is another guess about a cluster we can't see — the same kind of guess that has now produced three rounds of findings on this one knob.

Materializing costs nothing here because **the design already can't stream**. Both train loops do it themselves:

```python
df = normalize_pandas_for_lightgbm(shard.materialize().to_pandas())  # lgb.py:31
df = shard.materialize().to_pandas()                                  # xgb.py:20
```

A lightgbm `Dataset` / xgboost `DMatrix` is built from the whole local partition; there is no streaming ingest to give up. So this changes *when* execution happens, not *whether* the data ends up resident — and it removes the failure class rather than making it less likely.

For what it's worth, this matches how the first-party integrations behave: none of them computes a share from cluster totals. `lightgbm.dask` reads `client.ncores()` and pops a user `num_threads` with a warning; `xgboost.dask` reads `worker.state.nthreads`; `xgboost.spark` reads `spark.task.cpus` and lists `n_jobs`/`nthread` as unsupported. They all read a per-worker number the scheduler already set. Ray has none to read, which is why `xgboost_ray` had to invent one — and why it bounded it at both ends.

## Nits

- `test_workers_get_a_share_of_the_cluster_cpus:200` — kept as is, with a comment saying why: the CI cluster is a single node, so `min_node_cpus == cluster_cpus` and those expectations are still right. The multi-node behaviour is covered by a new parametrized test that monkeypatches `ray.nodes` (even, heterogeneous, and a dead node that mustn't drag the bound down), since a single-node cluster can't express it.
- `worker_n_jobs` coverage — restored as `test_worker_n_jobs_is_capped_by_the_assigned_cpus`, which calls it inside a `@ray.remote` with `num_cpus=2` rather than asserting on `model_` after a fit. The `== 2` assertion at `:72` stays; it correctly pins the restore.

## One thing worth knowing beyond this PR

Your round-3 note that `--timeout=600` doesn't reach users is right, and it doesn't reach CI either. With the materialize disabled, the new hang test ran **21 minutes past a `--timeout=90`** before I killed it. pytest-timeout's default raises from a SIGALRM handler, which only runs when the interpreter regains control — and the main thread is parked in ray's C++ core worker:

```
ray/_private/worker.py:1004, in get_objects
    ] = self.core_worker.get_objects(
+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++
```

That dump is from the same test with `method="thread"`, which fires correctly (45s, as configured). The new test uses the thread method for that reason.

`pytest.yml:98` runs `--timeout=600` with the comment "keeps a hung ray test from burning the job's full 6h budget" — as written that mitigation can't fire on this class of hang. `--timeout-method=thread` would make it real. Left out of this PR as unrelated; happy to send it separately.

## Verification

`22 passed` in 240s on ray 2.56.1 / lightgbm 4.6.0 / xgboost 3.4.1, python 3.13 (was 17). Plus the two-node end-to-end above, and the negative control: with `dataset.materialize()` commented out, the new test hangs — so it's a real regression guard, not a test that passes either way.

jmoralez (approved):

PR #732, opened 2026-09-04
fix: maintain Expanding*/EWM state in TimeSeries.update

### LINKED ISSUE #726: [BUG] `update()` doesn't maintain Expanding*/EWM accumulator state
`TimeSeries.update()` appends the new observations to `self.ga` but never advances the lag transforms' per-group state (`stats_`). Transforms that recompute from the stored array on every call are fine; the ones that carry a running accumulator are left stale, which shows up two ways.

### Symptom 1 — silently wrong features

The accumulator ends up behind by exactly `k`, where `k` is the number of new timestamps passed to `update()`. It happens with `k = 1`, so any `update()` at all is enough.

```python
import numpy as np, pandas as pd
from mlforecast.callbacks import SaveFeatures
from mlforecast.lag_transforms import ExpandingMin

T = 12
ds = pd.date_range("2020-01-01", periods=T, freq="D")
y = np.full(T, 100.0)
y[T - 2] = 1.0                      # the observation that gets skipped
df = pd.DataFrame({"unique_id": "a", "ds": ds, "y": y})

def expanding_min(data, update=None):
    fcst = MLForecast(
        freq="D", models=[LinearRegression()], lags=[1],
        lag_transforms={1: [ExpandingMin()]},
    )
    fcst.fit(data, static_features=[])
    if update is not None:
        fcst.update(update)
    cb = SaveFeatures()
    fcst.predict(1, before_predict_callback=cb)
    return cb.get_features()["expanding_min_lag1"].item()

print(expanding_min(df))                                              # 1.0    correct
print(expanding_min(df[df.ds < ds[-1]], update=df[df.ds == ds[-1]]))  # 100.0  wrong
```

The running minimum misses the only low value in the series. The same setup on `ExpandingMean` gives `100.0` instead of `91.75`.

How far behind, measured on a 20-step series with `ExpandingMean` (reading `ts.transforms["expanding_mean_lag1"]._core_tfm.stats_[0, 0]` after `predict`):

| | observations consumed |
|---|--:|
| `fit(full)` | 19 |
| `fit(T-1)` + `update(1 timestamp)` | 18 |
| `fit(T-2)` + `update(2 timestamps)` | 17 |
| `fit(T-3)` + `update(3 timestamps)` | 16 |
| `fit(T-5)` + `update(5 timestamps)` | 14 |

On a 20-point random series, `update(3)` moves `ExpandingStd` from 15.94 to 16.73 (~5%) and `ExpandingMean` from 48.79 to 48.84 — small enough to pass unnoticed, unlike the `ExpandingMin` case above.

### Symptom 2 — crash when `update()` adds a new series

`stats_` keeps its original row count while the group count grows:

```python
import numpy as np, pandas as pd
from mlforecast.lag_transforms import ExpandingMean

T = 10
ds = pd.date_range("2020-01-01", periods=T, freq="D")
df = pd.concat(
    [pd.DataFrame({"unique_id": u, "ds": ds, "y": np.arange(1.0, T + 1) + i})
     for i, u in enumerate(("a", "b"))],
    ignore_index=True,
)
fcst = MLForecast(
    freq="D", models=[LinearRegression()], lags=[1],
    lag_transforms={1: [ExpandingMean()]},
)
fcst.fit(df, static_features=[])
fcst.update(pd.DataFrame({
    "unique_id": ["a", "b", "c"],
    "ds": [ds[-1] + pd.Timedelta(days=1)] * 3,
    "y": [11.0, 12.0, 99.0],
}))
fcst.predict(1)
# ValueError: operands could not be broadcast together with shapes (2,) (3,) (2,)
```

### Root cause

`stats_` is primed by `transform()` at fit, consuming indices `0 .. N-lag-1`. At predict, `update()` advances it by exactly one observation, reading `ga._index_from_end(lag - 1)`. `TimeSeries.update()` appends `k` values in between and advances nothing, so those `k` observations are never folded in — only the single one the first predict step happens to read. The new-series case is the same gap: nothing allocates the extra `stats_` row.

### Affected

Every transform whose `coreforecast` inner carries `stats_`:

- `ExpandingMean`, `ExpandingStd`, `ExpandingMin`, `ExpandingMax`
- `ExponentiallyWeightedMean`

Not affected (stateless — each `update()` recomputes from the stored array): `Lag`, all `Rolling*`, all `SeasonalRolling*`, `RollingQuantile`, `ExpandingQuantile`.

### Suggested direction

`TimeSeries.update()` needs to advance the stateful transforms over the values it appends, rather than leaving them to be advanced once per `predict` step. Note that calling `update()` `k` times after a bulk `append_several` will not work — `_index_from_end` would read the same last value each time — so the values have to be folded in one timestamp at a time, or `coreforecast` needs a bulk-update entry point. The new-series case additionally needs `stats_` grown (and permuted, if the id ordering changes) to match the new group count.

For reference, the pooled engine had the identical new-series crash; the fix there was a `remap_buckets` hook that permutes and extends the inner state whenever the bucket vocabulary grows, seeding new buckets with the state of one that has only seen empty cells.

### Environment

mlforecast `main` @ 6f6d46d (also 1.1.0), coreforecast 0.0.18, numpy 2.4.1, pandas 2.3.3, Python 3.12.14.

jmoralez:
Closes #726.

## Problem

`Expanding*` and `ExponentiallyWeightedMean` carry a running accumulator (`stats_`) instead of recomputing from the stored array. coreforecast advances it by exactly one observation per `update` call, reading the value `lag` positions from the end, and relies on the serie having grown by one in between. `TimeSeries.update` appends `k` values and advances nothing, so:

- the accumulator falls behind by `k`, silently. With `ExpandingMin` on a series whose only low value is in the appended range, the running minimum misses it entirely (`100.0` instead of `1.0`).
- an update that introduces a new serie leaves `stats_` with fewer rows than groups, and the next `predict` raises `ValueError: operands could not be broadcast together`.

This also affects `DistributedMLForecast.update` and the conformal transfer backtest, which advances a `TimeSeries` one observation at a time.

## Fix

`TimeSeries._advance_lag_transform_states` folds the appended values into the accumulators one timestamp at a time, over a per-serie window of the last `max_lag` stored values followed by the new ones, so each step's read at `lag` from the end lands on the intended value. Series that run out of appended values get their `stats_` rows restored after each step, since coreforecast advances every group — that's what makes ragged updates work.

New series are handled after the fold: a copy of each transform is primed with `transform` over just that serie's values, and old and fresh rows are scattered into their positions in the grown array. That covers the reordering too, when a new id sorts into the middle of the existing ones.

Recomputing the state from scratch isn't an option — `keep_last_n` has already dropped the history the accumulator summarizes.

`_stateful_core_tfms` on the transforms reports which inner coreforecast transforms carry state; `Combine` covers both sides, `Offset` inherits it, and the pooled transforms report nothing since they never run a `transform` pass (they recompute from `_ts_aggs`, which was already correct across updates).

## Also fixed

These transforms reported `update_samples=1`, but each update reads the value `lag` back, so `keep_last_n` was inferred too small. At a lag above the largest plain lag the trim left too little history and every predicted value came out null:

```python
MLForecast(freq="D", models=[...], lags=[1], lag_transforms={3: [ExpandingMean()]})
# keep_last_n=1, expanding_mean_lag3 is NaN at every predict step
```

They now report the configured lag. This means `Expanding*`/EWM at a lag above the largest plain lag retain more history than before.

## Tests

New `tests/test_update_state.py`, 17 cases over pandas and polars: bulk updates (k = 1, 2, 5), sequential single-timestamp updates, ragged updates, a new serie sorting into the middle, and both repros from the issue. The equivalence tests compare `stats_` against a full fit *and* the full feature frames under a shared set of models, so the stateless transforms are covered too.

`test_keep_last_n_for_built_in_lag_transforms` expected `keep_last_n == 4` for `{1: [RollingMean(1), RollingMean(4)], 5: [ExpandingMean()]}`; that expectation encoded the bug above and is now 5.

https://claude.ai/code/session_01MXQFsD8TDT7L4cv8dLtSjg


nasaul (commented):
Nice catch on the root cause, and the one-timestamp-at-a-time fold is the right
shape. One blocking issue before this goes in, plus a few non-blocking notes.

## Blocking: NaN poisoning when stored history is shorter than the transform lag

`_advance_lag_transform_states` folds `counts[i]` values per serie
unconditionally. That's correct under the invariant *"after a fit on `M` values
the accumulator has consumed `M - L`"* — but when `M < L` a fit consumes **0**,
not a negative number. So the fold over-counts by `L - M` steps and reads those
steps out of the NaN pad at core.py#L491-L494. NaN is absorbing for every
stateful core, so the serie's feature is `nan` from then on and `predict` raises
on any model that rejects NaN.

Two reachable triggers, both reproduced against this branch:

**1. A new serie added by `update` with fewer observations than the lag, then
updated again.** No exotic flags — `test_update_with_new_series_matches_full_fit`
already walks this path, but at lag 1, where `L - M <= 0` makes it benign.

```python
# lag_transforms={4: [ExpandingMean()]}, fit on a, b (20 obs each)
fcst.update(frame("c", [1., 2.]))                       # new serie, 2 obs < lag 4
fcst.update(frame("c", [3., 4., 5.], start="2020-01-03"))

# expanding_mean_lag4 stats_
#   this branch: [16. 136.  16. 152.   3.  nan]   <- c poisoned, n is 3
#   full fit:    [16. 136.  16. 152.   1.   1.]
#   main:        [16. 136.  16. 152.]             <- c missing (the #726 shape bug)
```

**2. A user-supplied `keep_last_n` below the transform lag.**

```python
# lag_transforms={5: [ExpandingMean(), ExponentiallyWeightedMean(alpha=0.3)]}
fcst.fit(hist, static_features=[], keep_last_n=2)
fcst.update(tail)   # 3 timestamps

#   this branch: [23. nan  23. nan]   ewm [nan   nan  ]
#   full fit:    [23. 276.  23. 299.] ewm [20.67 21.67]
#   main:        [20. 210.  20. 230.] ewm [17.67 18.67]  <- stale, but finite
```

Both cases were already wrong on `main`, so this isn't strictly a regression —
but it turns "stale" into "NaN forever", which is a harder failure for models
that accept NaN natively and would otherwise have kept training.

### Suggested fix

Extend the existing restore machinery with a per-core "not yet started" mask —
the steps to skip are exactly those whose read position lands in the pad:

```python
stored_lens = np.diff(prev_ga.indptr)[existing]   # hoist above the `j` loop
...
for core in cores:
    skip = done | (j < core.lag - stored_lens)     # replaces `done`
    kept = core.stats_[skip].copy()
    core.update(step)
    core.stats_[skip] = kept
```

Per-core because `lag` varies across `cores`. I derived this but haven't tested
it, so treat it as a direction rather than a patch.

Worth noting the ceiling on case 2: skipping restores parity with `main` (stale,
finite) but not correctness — `keep_last_n=2` already destroyed the history the
accumulator summarizes. A `keep_last_n` below a stateful transform's lag might
deserve a warning at `fit` time, separately from this PR.

A regression test for a serie whose stored length is below the lag would be a
good addition to `test_update_state.py` — the existing new-serie tests all sit
at lag 1.

## Non-blocking

- `Offset.update_samples` returns `self.tfm.update_samples + self.n`, but
  `_set_core_tfm(lag + n)` already bakes the offset into the inner core's lag —
  so `n` is counted twice now that the inner property returns `core.lag`.
  `_pooled_retention` right below it has the correct comment about exactly this.
  Over-retention only, never under, and it predates this PR for `Rolling*`.
- `np.ascontiguousarray(context[:, j : j + max_lag]).ravel()` — `ravel()` on a
  non-contiguous slice already returns a contiguous copy, so the wrapper is
  redundant.
- Not caused by this PR: `forecast_fitted_values(h > 1)` raises
  `Input X contains NaN` with any `Expanding*` transform, identically on `main`
  and here — forecast.py#L1078 primes the cores after `keep_last_n` has trimmed.
  Worth its own issue.

nasaul (approved):
LGTM

PR #734, opened 2026-09-07
docs: fix documented defaults that disagree with the signatures

VenishPaneliya:
Four docstrings state defaults that don't match their signatures. Docs only, no behaviour change.

### `as_numpy` documented as `True`, actually `False`

`MLForecast.preprocess`, `MLForecast.fit` and `MLForecast.cross_validation` all declare:

```python
as_numpy: bool = False,
```

but each documents it as:

```
as_numpy (bool): Cast features to numpy array. Defaults to True.
```

So the docs say features are cast to a numpy array by default when they aren't. `MLForecast.predict` is the one that's already right — it takes `Optional[bool] = None` and documents "Defaults to None" — so only these three needed changing.

### `DistributedMLForecast.freq` documented as optional, actually required

```python
def __init__(
    self,
    models,
    freq: Freq,          # required, no default
    ...
```

```
freq (str or int, optional): Pandas offset alias, e.g. 'D', 'W-THU' or integer denoting the frequency of the series. Defaults to None.
```

`freq` has no default, so following the docs and omitting it raises `TypeError`. `MLForecast.__init__` documents the same required `freq: Freq` without "optional" and without a "Defaults to" clause:

```
freq (str or int or pd.offsets.BaseOffset): Pandas offset, pandas offset alias, e.g. 'D', 'W-THU' or integer denoting the frequency of the series.
```

so this just brings the distributed class in line with its sibling. The `models` entry immediately above it already follows that convention.

Checked with `ruff check` and `ruff format --check` on both changed files — clean.

### DIFF
--- a/mlforecast/distributed/forecast.py
+++ b/mlforecast/distributed/forecast.py
@@ -81,7 +81,7 @@ class DistributedMLForecast:
 
         Args:
             models (regressor or list of regressors): Models that will be trained and used to compute the forecasts.
-            freq (str or int, optional): Pandas offset alias, e.g. 'D', 'W-THU' or integer denoting the frequency of the series. Defaults to None.
+            freq (str or int): Pandas offset alias, e.g. 'D', 'W-THU' or integer denoting the frequency of the series.
             lags (list of int, optional): Lags of the target to use as features. Defaults to None.
             lag_transforms (dict of int to list of functions, optional): Mapping of target lags to their transformations. Defaults to None.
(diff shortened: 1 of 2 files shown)


jmoralez:
Thanks for the fixes @VenishPaneliya! Can you please sign the CLA?

VenishPaneliya:
Thanks for taking a look! I'm getting the CLA signed now and will confirm here once the check goes green.

jmoralez (approved):
Thank you for your contribution!

PR #735, opened 2026-09-07
fix: preserve target transform state in cross-validation

### LINKED ISSUE #733: Target transforms corrupt conformal calibration with refit=False cross-validation
### What happened + what I expected

When `MLForecast.fit(..., prediction_intervals=...)` is used together with a learned target transform such as `LocalStandardScaler`, conformal calibration can produce very large errors even when the underlying estimator predicts the series exactly.

I expected the fitted estimator and target-transform state to remain consistent throughout the `refit=False` cross-validation used to compute conformity scores.

The final point forecasts are correct, but the intervals are much too wide. In the deterministic example below, the untransformed model has the expected zero-width interval, while adding `LocalStandardScaler` changes the 80% interval to approximately `[54, 146]` around a point forecast of `100`.

### Reproduction

```python
import numpy as np
import pandas as pd
from mlforecast.target_transforms import LocalStandardScaler
from mlforecast.utils import PredictionIntervals

n = 100
df = pd.DataFrame(
    {
        "unique_id": "series",
        "ds": pd.date_range("2025-01-01", periods=n, freq="D"),
        "y": np.arange(n, dtype=float),
        "trend": np.arange(n, dtype=float),
    }
)
future = pd.DataFrame(
    {
        "unique_id": "series",
        "ds": pd.date_range("2025-04-11", periods=2, freq="D"),
        "trend": [100.0, 101.0],
    }
)

for name, transforms in [
    ("none", None),
    ("LocalStandardScaler", [LocalStandardScaler()]),
]:
    fcst = MLForecast(
        models={"linear": LinearRegression()},
        freq="D",
        target_transforms=transforms,
    )
    fcst.fit(
        df,
        static_features=[],
        prediction_intervals=PredictionIntervals(n_windows=20, h=2),
    )
    print(name)
    print(fcst.predict(h=2, X_df=future, level=[80]).to_string(index=False))
```

Output:

```text
none
unique_id         ds  linear  linear-lo-80  linear-hi-80
   series 2025-04-11   100.0         100.0         100.0
   series 2025-04-12   101.0         101.0         101.0

LocalStandardScaler
unique_id         ds  linear  linear-lo-80  linear-hi-80
   series 2025-04-11   100.0     54.334598    145.665402
   series 2025-04-12   101.0     54.831148    147.168852
```

### Suspected cause

`MLForecast.fit` computes conformity scores through `cross_validation(refit=False)`. The first window fits the estimator and target transforms together. For later windows, cross-validation calls `predict(new_df=train)`. That creates a new `TimeSeries` and runs the target transforms' `fit_transform` on the expanded history, while reusing the estimator fitted against the first window's transformed target scale.

The transform and estimator therefore no longer share the same fitted state. Inverting predictions with the newly fitted transform creates artificial calibration errors, which then inflate the conformal intervals.

This affects learned scaling transforms (`LocalStandardScaler`, `LocalMinMaxScaler`, and `LocalRobustScaler` in my checks). Compositions can also be affected; for example, `Differences(...)` followed by `LocalStandardScaler()` fails on a nonlinear trend.

### Possible approach

For `cross_validation(refit=False)`, could the implementation preserve the transform state fitted with the estimator and advance it between windows with the newly observed targets using the target transforms' native update path (`TimeSeries.update`)?

That would retain frozen-estimator semantics while keeping stateful transforms synchronized. A full per-window refit of both estimator and transforms is also correct, but has different semantics and cost.

It may also be useful to fail explicitly for a custom stateful transform that cannot be updated under `refit=False`, or allow a documented refit strategy.

### Versions

```text
Python 3.14.7
mlforecast 1.1.0
numpy 2.4.6
pandas 2.3.3
scikit-learn 1.9.0
```

ramirocrc:
Closes #733.

## What changed

- Keep the target-transform state fitted with each frozen estimator during
  `cross_validation(refit=False)` and periodic-refit cross-validation.
- Advance target state independently from `TimeSeries`, using only observations
  added since the previous window.
- Add an internal `_transform_history` contract that applies learned transform
  parameters to a bounded historical window while rebuilding runtime state used
  for prediction and fitted-value inversion.
- Build a scratch `TimeSeries` for each non-refit window instead of mutating the
  fitted instance.
- Warm only prediction state when `fitted=False`; materialize training features
  only when cross-validation fitted values are requested.
- Share update ID normalization, group sizing, new-series detection, and last-date
  advancement with `TimeSeries.update`.

## Why

The first cross-validation window fits the estimator and target transforms
together. Later non-refit windows previously called `predict(new_df=train)`, which
refit target transforms on each window while reusing the estimator from the first
window. Learned transforms such as `LocalStandardScaler` therefore moved to a new
target scale without refitting the estimator, creating artificial conformity
errors and severely inflated conformal intervals.

Advancing the fitted transform state in place fixes the scale mismatch for aligned
expanding panels, but changes other cross-validation semantics. In particular, it
mutates the fitted `TimeSeries`, breaks Polars categorical IDs, stops `input_size`
and `keep_last_n` from bounding window state, and leaves fitted-value inversion on
the wrong transform state.

The scratch-window implementation keeps the estimator and learned transform
parameters frozen while rebuilding exactly the feature and target runtime state
required by each bounded cross-validation window.

## Behavior covered

- pandas and Polars, including Polars categorical IDs
- scaling alone and scaling composed with `Differences`
- auto-differencing transforms without relearning their selected differences
- `refit=False` and periodic integer refits
- overlapping windows
- `fitted=False` and `fitted=True`
- `input_size` and `keep_last_n`
- pooled lag transforms
- unaligned series histories
- collision-safe internal columns and explicit unknown-series errors
- restoration of the fitted `TimeSeries` when prediction raises
- actionable failure for custom transforms without historical replay support

## Validation

- Ruff formatting and lint checks pass for all modified files.
- 186 core and target-transform tests pass.
- 35 focused frozen-model cross-validation tests pass.
- 7 pooled target-transform tests pass.

### DIFF
--- a/mlforecast/core.py
+++ b/mlforecast/core.py
@@ -7,6 +7,7 @@ import reprlib
 import warnings
 from collections import Counter, OrderedDict
 from contextlib import contextmanager
+from dataclasses import dataclass
 from pathlib import Path
 from typing import (
     Any,
@@ -251,6 +252,197 @@ def _to_native_index(values, *, df):
     if isinstance(df, pd.DataFrame):
         return pd.Index(values)
     return pl_Series(values)
+
+
+@dataclass
+class _MatchedUpdate:
+    df: DataFrame
+    uids: Any
+    matched_ids: Any
+
+
+@dataclass
+class _PreparedUpdate(_MatchedUpdate):
+    """Backend-normalized inputs shared by all update paths."""
+
+    id_counts: DataFrame
+    sizes: DataFrame
+    new_groups: Any
+
+
+def _match_update_data(
+    df: DataFrame,
+    *,
+    uids: Any,
+    id_col: str,
+    time_col: str,
+) -> _MatchedUpdate:
+    """Match update IDs to the fitted backend and sort the rows."""
+    native_uids = _index_to_series(uids)
+    native_uids, matched_ids = ufp.match_if_categorical(native_uids, df[id_col])
+    df = ufp.copy_if_pandas(df, deep=False)
+    df = ufp.assign_columns(df, id_col, matched_ids)
+    df = ufp.sort(df, by=[id_col, time_col])
+    return _MatchedUpdate(df=df, uids=native_uids, matched_ids=matched_ids)
+
+
+def _prepare_update_data(matched: _MatchedUpdate, *, id_col: str) -> _PreparedUpdate:
+    """Compute group sizes and detect series introduced by an update."""
+    id_counts = ufp.counts_by_id(matched.df, id_col)
+    try:
+        sizes = ufp.join(matched.uids, id_counts, on=id_col, how="outer_coalesce")
+    except (KeyError, ValueError):
+        # pandas raises key error, polars before coalesce raises value error
+        sizes = ufp.join(matched.uids, id_counts, on=id_col, how="outer")
+    sizes = ufp.fill_null(sizes, {"counts": 0})
+    sizes = ufp.sort(sizes, by=id_col)
+    new_groups = ~ufp.is_in(sizes[id_col], matched.uids)
+    return _PreparedUpdate(
+        df=matched.df,
+        uids=matched.uids,
+        matched_ids=matched.matched_ids,
+        id_counts=id_counts,
+        sizes=sizes,
+        new_groups=new_groups,
+    )
+
+
+def _updated_last_dates(
+    df: DataFrame,
+    *,
+    uids: Any,
+    sizes: DataFrame,
+    last_dates: Any,
+    id_col: str,
+    time_col: str,
+) -> Any:
+    """Advance each fitted series' last date from a prepared update."""
+    current_col = "_mlforecast_current_last_date"
+    while current_col in df.columns:
+        current_col = f"_{current_col}"
+    updated = ufp.group_by_agg(df, id_col, {time_col: "max"})
+    updated = ufp.join(sizes, updated, on=id_col, how="left")
+    current = type(df)({id_col: uids, current_col: last_dates})
+    updated = ufp.join(updated, current, on=id_col, how="left")
+    updated = ufp.fill_null(updated, {time_col: updated[current_col]})
+    updated = ufp.sort(updated, by=id_col)
+    values = ufp.cast(updated[time_col], last_dates.dtype)
+    return _to_native_index(values, df=df)
+
+
+@dataclass
+class _TargetTransformState:
+    """Target-only state advanced between frozen-model CV windows."""
+
+    transforms: List[TargetTransform]
+    uids: Any
+    last_dates: Any
+    id_col: str
+    time_col: str
+    target_col: str
+    target_dtype: np.dtype
+
+    @classmethod
+    def from_time_series(cls, ts: "TimeSeries") -> "_TargetTransformState":
+        """Copy only state required to advance and replay target transforms."""
+        if ts.target_transforms is None:
+            raise ValueError("No fitted target transforms are available.")
+        return cls(
+            transforms=copy.deepcopy(ts.target_transforms),
+            uids=copy.deepcopy(ts.uids),
+            last_dates=copy.deepcopy(ts.last_dates),
+            id_col=ts.id_col,
+            time_col=ts.time_col,
+            target_col=ts.target_col,
+            target_dtype=ts.ga.data.dtype,
+        )
+
+    def new_observations(self, train: DataFrame) -> DataFrame:
+        """Return rows in ``train`` that are newer than this state."""
+        last_date_col = "_mlforecast_last_date"
+        while last_date_col in train.columns:
+            last_date_col = f"_{last_date_col}"
+        last_dates = type(train)(
+            {self.id_col: self.uids, last_date_col: self.last_dates}
+        )
+        updates = ufp.join(train, last_dates, on=self.id_col, how="left")
+        unknown = (
+            nw.from_native(updates[last_date_col], series_only=True)
+            .is_null()
+            .to_native()
+        )
+        if unknown.any():
+            unknown_ids = ufp.filter_with_mask(updates[self.id_col], unknown)
+            raise ValueError(
+                "Can not update target_transforms with new series: "
+                f"{list(unknown_ids)!r}."
+            )
+        updates = ufp.filter_with_mask(
+            updates, updates[self.time_col] > updates[last_date_col]
+        )
+        return ufp.drop_columns(updates, last_date_col)
+
+    def update(self, df: DataFrame) -> None:
+        """Advance target transforms without creating feature state."""
+        validate_format(df, self.id_col, self.time_col, self.target_col)
+        matched = _match_update_data(
+            df,
+            uids=self.uids,
+            id_col=self.id_col,
+            time_col=self.time_col,
+        )
+        prepared = _prepare_update_data(matched, id_col=self.id_col)
+        if prepared.new_groups.any():
+            raise ValueError("Can not update target_transforms with new series.")
+        df = prepared.df
+        values = df[self.target_col].to_numpy().astype(self.target_dtype, copy=False)
+        has_grouped_transforms = any(
+            isinstance(tfm, _BaseGroupedArrayTargetTransform) for tfm in self.transforms
+        )
+        if has_grouped_transforms:
+            indptr = np.append(0, prepared.sizes["counts"]).cumsum()
+        for tfm in self.transforms:
+            if isinstance(tfm, _BaseGroupedArrayTargetTransform):
+                ga = tfm.update(GroupedArray(values, indptr))
+                df = ufp.assign_columns(df, self.target_col, ga.data)
+            else:
+                df = tfm.update(df)
+            values = df[self.target_col].to_numpy()
+
+        self.last_dates = _updated_last_dates(
+            df,
+            uids=prepared.uids,
+            sizes=prepared.sizes,
+            last_dates=self.last_dates,
+            id_col=self.id_col,
+            time_col=self.time_col,
+        )
+
+    def replay_history(
+        self, df: DataFrame, store_fitted: bool
+    ) -> Tuple[DataFrame, List[TargetTransform]]:
+        """Replay a bounded history with frozen fitted parameters."""
+        transformed = ufp.sort(df, by=[self.id_col, self.time_col])
+        transformed = ufp.copy_if_pandas(transformed, deep=False)
+        uids, _, values, indptr, _ = ufp.process_df(
+            transformed, self.id_col, self.time_col, self.target_col
+        )
+        if list(uids) != list(self.uids):
+            raise ValueError("History must contain exactly the fitted series.")
+        if values.ndim == 2:
+            values = values[:, 0]
+        transforms = copy.deepcopy(self.transforms)
+        ga = GroupedArray(values, indptr)
+        for tfm in transforms:
+            if hasattr(tfm, "store_fitted"):
+                tfm.store_fitted = store_fitted
+            if isinstance(tfm, _BaseGroupedArrayTargetTransform):
+                ga = tfm._transform_history(ga)
+                transformed = ufp.assign_columns(transformed, self.target_col, ga.data)
+            else:
+                transformed = tfm._transform_history(transformed)
+                ga.data = transformed[self.target_col].to_numpy()
+        return transformed, transforms
 
 
 class TimeSeries:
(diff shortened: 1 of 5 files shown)


nasaul:
Nice catch on the root cause — the diagnosis in #733 is exactly right, and advancing frozen transform state via `TimeSeries.update` matches the existing pattern in `_compute_recursive_fitted_values_on_demand` (`forecast.py:1091`), so the approach fits the library.

The core fix works: on an aligned panel, non-refit CV max error drops from 0.67 to ~1e-13.

I pushed four failing regression tests to this branch (`tests/test_forecast.py`) for the issues below. Three of them pass on `main` and fail here; the fourth fails identically on both.

One structural note before the specifics. The precedent at `forecast.py:1091` advances a **scratch** `TimeSeries` built for the loop; this PR advances `self.ts` in place. Three of the four issues fall out of that difference, so a CV-scoped instance may be cheaper than patching them individually.

---

### 1. polars `Categorical` id column breaks `cross_validation(refit=False)`

`self.update(updates)` routes through `TimeSeries.update`, which normalizes ids with `utilsforecast.processing.match_if_categorical`. Its polars branch recasts inside a **temporary** `pl.StringCache()` that closes on exit, then `update` writes the result back to `self.uids`. `cutoffs` was built by `ufp.backtest_splits` from the original `df`, so the join at `forecast.py:2063` compares categoricals from two different caches:

```
polars.exceptions.StringCacheMismatchError: cannot compare categoricals coming
from different sources, consider setting a global StringCache.
```

Same for `GlobalSklearnTransformer`, `AutoDifferences`, `LocalBoxCox`. Plain-string polars and pandas `category` are unaffected — the new parametrization uses `pl.from_pandas`, which gives `pl.String`.

Test: `test_cv_no_refit_polars_categorical_id`.

### 2. The fix doesn't reach `cross_validation_fitted_values()`

`_compute_fitted_values` runs at `forecast.py:2011`, before the `self.ts = prediction_ts` restore at 2025. So it inverse transforms with the deep copy whose transforms were refit on the expanded window, while `models_` is still frozen on window 0 — the same mismatch this PR fixes for predictions.

n=100, `LocalStandardScaler`, `n_windows=5, h=2, step_size=1, refit=False, fitted=True`:

| fold | prediction max err | fitted-value max err |
|---|---|---|
| 0 | 0.0 | 0 |
| 1 | 0.0 | 190 |
| 2 | 0.0 | 391 |
| 3 | 0.0 | 601 |
| 4 | 0.0 | 823 |

The deepcopy does real work — it stops `preprocess` from clobbering the state the prediction path needs — but the fitted values themselves are unchanged. The added test fails byte-identically on `main` and on this branch (386/480 mismatched, max abs diff 4.149), so this is a pre-existing gap rather than a regression. Flagging it because the PR's framing implies it is covered, and the new `fitted=True` parametrization asserts only on `cv["LinearRegression"]`.

Test: `test_cv_no_refit_fitted_values_preserve_target_transform_state`.

### 3. `input_size` and `keep_last_n` stop bounding the state

`TimeSeries.update` ends in `self.ga.append_several(...)` — append only, no `_apply_keep_last_n`, no truncation. The old `predict(new_df=train)` rebuilt state from exactly the window `backtest_splits` handed over, which honors `input_size`. Now window 0's history is kept and only deltas are appended, so a rolling-window CV silently becomes expanding. `keep_last_n` likewise stops capping retained history, so memory grows across windows.

The test uses `Differences([1])` deliberately: it is parameter free, so refitting it per window is *correct*, which makes `predict(new_df=train)` a valid reference and isolates the state bound from the transform-freezing question. `ExpandingMean` makes the retained history observable. Result: 3 of 4 windows mismatch, max abs diff 0.143.

Test: `test_cv_no_refit_honors_input_size`.

### 4. `_last_date` collides, and the `>` filter swallows unknown ids

`"_last_date"` is hardcoded with no uniqueness check. A user frame that already has that column makes the pandas merge suffix both sides to `_last_date_x` / `_last_date_y`, and `updates[time_col] > updates["_last_date"]` raises `KeyError: '_last_date'`.

Separately, the join is `how="left"` from `train`, so an id in `train` that isn't in `self.ts.uids` gets null for `_last_date`, `time > NaT` is `False`, and every row of that series is filtered out silently. That routes around `TimeSeries.update`'s explicit guard:

```python
if new_groups.any():
    if self.target_transforms is not None:
        raise ValueError("Can not update target_transforms with new series.")
```

I did not test the second half — `backtest_splits` can't produce a mid-panel new series, so reaching it needs an internals-level fixture.

Test: `test_cv_no_refit_last_date_column_collision`.

### 5. Restore isn't exception-safe

If `preprocess` at `forecast.py:1990` raises, line 2025 never runs and `self.ts` is left pointing at the deep copy with refit transforms — and the exception propagates out of `cross_validation`, so the caller keeps a wrong-state object. `try/finally` costs nothing. Untested.

---

### Nits

- `forecast.py:1972-1978, 2025` — the `prediction_ts = None` sentinel and the `if prediction_ts is not None:` guard both disappear if you save and restore unconditionally (restoring to itself is a no-op when there are no transforms). That also lets the two back-to-back `if self.ts.target_transforms is not None:` checks at 1973 and 1978 merge.
- `forecast.py:2042-2047` — the trailing comma after `self.ts.last_dates` forces black to explode a one-line dict into six. Dropping it collapses the block to three lines.

ramirocrc:
Thanks for the review @nasaul !

  I’ve pushed a follow-up that:

  - Keeps target-transform state separate from TimeSeries.
  - Builds a scratch TimeSeries for each non-refit window.
  - Preserves input_size and keep_last_n.
  - Handles Polars categorical IDs and internal-column collisions.
  - Uses the correct transform state for fitted values.
  - Restores the original state safely if prediction fails.
  - Adds an internal history-replay contract for target transforms.

  I also added coverage for auto-differencing, pooled transforms, periodic refits, unaligned histories, unknown IDs, and the other cases you raised.

  When you have a chance, I’d appreciate another look. Thanks again!

mlforecast/core.py
```diff
@@ -253,6 +254,197 @@ def _to_native_index(values, *, df):
...
+        if prepared.new_groups.any():
+            raise ValueError("Can not update target_transforms with new series.")
+        df = prepared.df
+        values = df[self.target_col].to_numpy().astype(self.target_dtype, copy=False)
+        has_grouped_transforms = any(
+            isinstance(tfm, _BaseGroupedArrayTargetTransform) for tfm in self.transforms
+        )
+        if has_grouped_transforms:
+            indptr = np.append(0, prepared.sizes["counts"]).cumsum()
+        for tfm in self.transforms:
```
nasaul, line 404: **HIGH — `_TargetTransformState.update`'s transform loop crashes `AutoDifferences` when series need different numbers of differences, and the loop is dead code.**

`indptr` is built from `prepared.sizes["counts"]`, then each transform's `update` is called. For `AutoDifferences`, coreforecast's `AutoDifferences._update` assumes every group's tail has exactly `season_length` elements (`tail_indptr = np.arange(0, season_length * ga.indptr.size, season_length)`), which is false whenever `diffs_` differs across series.

Reproduced with two series with unaligned end dates, `AutoDifferences(2)`, `refit=False`:

```
ValueError: Last element of indptr must be equal to the size of data
  core.py:406 in _TargetTransformState.update -> target_transforms.py:261 -> coreforecast/scalers.py:420
```

This works on `main`. Worse, the loop has no effect for any shipped transform - `replay_history` rebuilds all transform state from the full `train` (`Differences._transform_history` -> `fit_transform`, `AutoDifferences._transform_history` resets `tails_`, `_BaseLocalScaler`/`GlobalSklearnTransformer` are frozen). The only lasting effect of `update()` is advancing `self.last_dates`. Deleting lines 404-410 fixes the crash and all 19 new tests still pass (verified).

mlforecast/core.py
```diff
@@ -253,6 +254,197 @@ def _to_native_index(values, *, df):
...
+        updates = ufp.join(train, last_dates, on=self.id_col, how="left")
+        unknown = (
+            nw.from_native(updates[last_date_col], series_only=True)
+            .is_null()
+            .to_native()
+        )
+        if unknown.any():
+            unknown_ids = ufp.filter_with_mask(updates[self.id_col], unknown)
+            raise ValueError(
+                "Can not update target_transforms with new series: "
```
nasaul, line 377: **HIGH — `refit=False` CV now hard-fails when a series first appears after the initial cutoff.**

`new_observations` raises `"Can not update target_transforms with new series"` for any id in `train` that isn't in the fit window's `uids`. Staggered panel starts are ordinary.

Reproduced with series `a` full-length, series `b` starting after the first cutoff, `target_transforms=[LocalStandardScaler()]`, `refit=False`:

```
ValueError: Can not update target_transforms with new series: ['b', 'b', 'b'].
```

`main` returns results for the same input. Two problems in one: the hard failure, and the message lists one entry per offending *row* rather than unique ids (needs `ufp.unique`/`set`).

mlforecast/target_transforms.py
```diff
@@ -38,6 +38,17 @@ def set_column_names(self, id_col: str, time_col: str, target_col: str):
     def update(self, df: DataFrame) -> DataFrame:
         raise NotImplementedError
 
+    def _transform_history(self, df: DataFrame) -> DataFrame:
```
nasaul, line 41: **MEDIUM — the `_transform_history` `NotImplementedError` default breaks every user-defined target transform that worked before.** (Same issue on the `_BaseGroupedArrayTargetTransform` default at line 75.)

Any `BaseTargetTransform` subclass that implements only `fit_transform`/`inverse_transform`/`update` - the documented contract - now dies in `cross_validation(refit=False)`.

Reproduced with a 5-line `Log1p(BaseTargetTransform)`:

```
NotImplementedError: Log1p must implement _transform_history to support cross_validation with frozen models.
```

Works on `main`. Defaulting `_transform_history` to `return self.fit_transform(...)` reproduces the old behavior exactly and matches what `Differences` already does at line 146.

mlforecast/forecast.py
```diff
@@ -2027,14 +2071,16 @@ def cross_validation(
...
                 X_df = None
-            y_pred = self.predict(
-                h=h,
-                before_predict_callback=before_predict_callback,
-                after_predict_callback=after_predict_callback,
-                new_df=train if not should_fit else None,
-                level=level,
-                X_df=X_df,
-            )
+            prediction_df = train if not should_fit and window_ts is None else None
```
nasaul, line 2074: **MEDIUM — conformal-interval calibration semantics now depend on whether `target_transforms` is configured.**

`prediction_df = train if not should_fit and window_ts is None else None` - `window_ts` is only built when target transforms exist, so with transforms `predict` takes the same-domain path (fit-window `_cs_df`, constant widths), and without them it still takes the `new_df`-driven transfer/`_frozen_backtest` path (per-window recalibration).

Measured on identical data, `n_windows=5, h=3, refit=False, level=[80]`, mean interval width by cutoff:

```
no target_transforms:  5.30, 4.67, 4.48, 1.48, 3.64   (recalibrated per window)
LocalStandardScaler:   5.30, 5.30, 5.30, 5.30, 5.30   (fit-window scores reused)
```

The frozen-score behavior is probably the intended one, but it should not be reachable only via the presence of a target transform.

mlforecast/target_transforms.py
```diff
@@ -122,6 +140,11 @@ def update(self, ga: GroupedArray) -> GroupedArray:
             core_ga = core_ga._with_data(transformed)
         return GroupedArray(transformed, ga.indptr)
 
+    def _transform_history(self, ga: GroupedArray) -> GroupedArray:
+        # Differences has no learned parameters. Rebuilding the scalers gives
+        # them tails aligned with the end of this history.
+        return self.fit_transform(ga)
```
nasaul, line 146: **LOW — `Differences._transform_history` delegates to `fit_transform`, which raises the internal `_ShortSeriesException`.**

`_fit_core` wraps that into a friendly `ValueError` naming the offending series; `replay_history` does not. A window bounded by `input_size` that is shorter than `sum(differences)` therefore surfaces a raw internal exception carrying a tuple of positional indices instead of series ids.

nasaul (commented):
Correctness review of the frozen-model CV rework. Five findings posted inline: two HIGH regressions against `main`, two MEDIUM behavior/compat changes, one LOW error-message issue.

### Looks right / verification

- `tests/test_forecast.py`, `tests/test_target_transforms.py`, `tests/test_pooled.py`: **653 passed, 2 skipped** on the PR head. The 19 new frozen-CV tests pass. The regressions below are coverage gaps, not test failures.
- **Numeric parity oracle:** 3 series, `lags=[1,2,7]`, `RollingMean(7)` + `ExpandingMean()`, `Differences([1])` (no learned parameters, so `main` and this PR should agree), `refit=False`, 6 windows, `h=4` -> **max abs difference `0.0`**. The new scratch-`TimeSeries` path reproduces the old feature and lag-transform state exactly.
- `AutoDifferences._transform_history` checked line-by-line against coreforecast's `AutoDifferences._transform` / `AutoSeasonalityAndDifferences.fit_transform`: step count, `_diffs_to_indptr` equivalence (`np.append(0, d.cumsum())` - the int64-vs-int32 indptr is accepted by the nanobind binding, verified empirically), `tails_` construction, and the `diffs.max() == 0` empty case all match.
- `_copy_config` covers every `TimeSeries.__init__` argument, and `_parse_transforms` deep-copies lag transforms, so the scratch window cannot clobber the fitted instance's lag-transform state. `_use_time_series` restores `self.ts` in `finally`. `has_non_refit_windows = n_windows > 1 and refit != 1` correctly treats `refit=True` as 1 and `refit=False`/`0` as non-refit.
- Spot-checked `input_size=30`, `keep_last_n=10`, `max_horizon=3`, and `fitted=True, max_horizon=3` against both `Differences([7])` and `LocalStandardScaler` - all clean.

PR #739, opened 2026-09-09
refactor: one writer for TimeSeries fit settings, copies for sub-steps

jmoralez:
Replaces the "declare every instance attribute once" approach (previous version of this PR) with a narrower change that targets the two things that actually cause trouble: a method silently overriding a setting another method established, and a second `TimeSeries` being assembled from a hand-written list of the first one's attributes. Public API and pickle layout are unchanged.

Two rules, both about ownership:

1. **`TimeSeries` is the only writer of its fit settings.** Other code passes them to `fit_transform`/`history_warmup` or asks for a clone.
2. **Sub-steps run on a copy.** A method that runs fit, preprocess or cross-validation as a step of something else does it on a copy of the forecaster, never on `self`.

### What changed

- `TimeSeries` gets class-level defaults for `as_numpy`, `max_horizon`, `_horizons` and `horizon_features_`, the settings that `_fit`-only instances (LightGBMCV, distributed partitions) and older pickles lack. The three `as_numpy = ...` write-backs (`MLForecast.fit`, `LightGBMCV.fit`, partition preprocess) and the `getattr(self, ..., default)` preserves in `history_warmup` go away.
- `fit_transform`/`history_warmup` take `horizon_features`; `MLForecast.preprocess`/`history_warmup` pass the resolved mapping instead of writing `ts.horizon_features_`.
- `TimeSeries._fit_settings()` is the single list of what a fresh instance needs; `_clone_cold()` and `_clone_warm(df)` are its only readers. `predict(new_df=...)` and the on-demand fitted-values path use them instead of rebuilding a `TimeSeries` by hand.
- Target transforms grow a `clone()` that returns an unfitted instance, since `_clone_cold()` needs one and deep-copying dragged the fitted state along. Every override builds from `type(self)` so a user subclass survives the clone, which the previous `copy.deepcopy` gave for free.
- `MLForecast._with_ts()` (shallow copy) hosts the sub-steps: the conformity-score cross-validation in `fit`, the transfer-conformal preprocess in `predict`, and `_frozen_backtest`. The five `_saved_*` locals, the `_saved_cs_df` swap and both `try/finally` blocks are gone. `fit(prediction_intervals=...)` no longer leaves `cv_models_`/`cv_fitted_values_` behind (new test).
- `MLForecast._from_ts()` replaces the construct-then-overwrite pattern in `load`, `from_cv` and `DistributedMLForecast.to_local`.
- `prediction_intervals`, `_cs_df` and `_cs_source_scales_` get class-level `None`, so the `hasattr` guards go and instances pickled before those fields existed still load.
- `predict`'s docstring states that on success the instance continues from `new_df`.

Reviewing with `git diff -w` is recommended: removing the `try/finally` around the forecast/interval step in `predict` dedents ~170 lines.

https://claude.ai/code/session_01S18rqZJqQERqNyQ3hj7k5h


mlforecast/core.py
```diff
@@ -1489,21 +1503,73 @@ def history_warmup(
...
+        """Arguments that warm a fresh instance the way this one was fit."""
+        settings: Dict[str, Any] = dict(
+            id_col=self.id_col,
+            time_col=self.time_col,
+            target_col=self.target_col,
+            static_features=self.static_features,
+            keep_last_n=self.keep_last_n,
+            weight_col=self.weight_col,
+            as_numpy=self.as_numpy,
+            horizon_features=copy.deepcopy(self.horizon_features_),
```
nasaul, line 1528: `horizon_features_` belongs in the class-level defaults added above.

The block at core.py:274-279 says exactly why those defaults exist:

> Instances that only went through `_fit` (LightGBMCV, distributed partitions) and older pickles fall through to the value `predict` assumed for them.

`horizon_features_` is in the same category — it's set in `__init__` (line 326) and written by `fit_transform`/`history_warmup`, never by `_fit` — and it first shipped in **v1.1.0**, the same release as `_horizons` (`git log -S horizon_features_ -- mlforecast/core.py` → `d835fb4`, tagged v1.1.0).

So a `TimeSeries` pickled by ≤ v1.0.2 lacks both: `_horizons` now falls through to the class default, but `self.horizon_features_` on this line raises `AttributeError`. That propagates to every `_clone_warm` caller — `predict(new_df=...)` (forecast.py:1519) and the on-demand fitted-values path (forecast.py:1083), which didn't touch the attribute before this PR.

One line next to the others fixes it:

```python
horizon_features_: Dict[int, List[str]] = {}
```

A mutable class default is safe here since every writer rebinds rather than mutates, and `__init__` sets an instance-level `{}` anyway. If you'd rather not have a shared dict on the class, `getattr(self, "horizon_features_", {})` on this line works too.

Worth adding `horizon_features_` to the pop list in `test_fit_only_instance_uses_class_defaults` — as written it covers the three defended attributes but not this one, so nothing catches the gap.

Note this isn't a regression for `predict(new_df=...)` specifically (main dereferenced it at forecast.py:1549 too), but the PR's stated goal of older pickles continuing to load isn't reached without it.

---
_Generated by [Claude Code](https://claude.ai/code)_

mlforecast/target_transforms.py
```diff
@@ -92,6 +113,9 @@ class Differences(_BaseGroupedArrayTargetTransform):
     def __init__(self, differences: Iterable[int]):
         self.differences = list(differences)
 
+    def _clone(self) -> "Differences":
```
nasaul, line 116: `_clone` hardcodes the class instead of using `type(self)`, which loses user subclasses.

`_clone_cold` (core.py:1548) replaces what used to be `copy.deepcopy(self.ts.target_transforms)` with `[tfm.clone() for tfm in ...]`. `deepcopy` preserved the concrete class; `clone()` doesn't, here and in `AutoDifferences` (221), `AutoSeasonalDifferences` (355), `AutoSeasonalityAndDifferences` (388) and `GlobalSklearnTransformer.clone` (488).

Concretely: someone who subclasses `Differences` to override `inverse_transform` gets a plain `Differences` inside the clone, so `predict(new_df=...)` and the transfer-conformal preprocess silently return wrongly back-transformed values — no error, just different numbers than `predict()` without `new_df` gives.

The inconsistency looks unintended, since `_BaseLocalScaler._clone` (437) gets it right:

```python
out = type(self).__new__(type(self))
out.scaler_factory = self.scaler_factory
```

`type(self)(...)` isn't quite enough for `Differences`/`AutoDifferences` on its own, because the subclass `__init__` signatures differ — the `_BaseLocalScaler` pattern (`__new__` plus copying the argument attributes) is the consistent fix.

`test_clone_is_unfitted_and_independent` does assert `type(cloned) is type(tfm)`, but only over the eight built-ins, where the hardcoded class happens to match. A single subclass in the parametrize list would fail it today.

Minor related note on line 439: `_BaseLocalScaler.take` changed from `copy.deepcopy(self)` to `self.clone()`, so it now carries only `scaler_factory` + `num_threads`. Correct for all four built-in local scalers, but it silently drops any other attribute a subclass sets in `__init__`.

---
_Generated by [Claude Code](https://claude.ai/code)_

nasaul (commented):
Reviewed the full diff and traced every call site of the new helpers. Two findings, both small fixes, left as inline comments:

1. **`mlforecast/core.py:1528`** — `_fit_settings` dereferences `horizon_features_`, which was left out of the class-level defaults this PR added for exactly this case. Pre-v1.1.0 pickles still raise `AttributeError` in `_clone_warm`.
2. **`mlforecast/target_transforms.py:116`** (and 221, 355, 388, 488) — `_clone`/`clone` build a hardcoded class rather than `type(self)`. Since `_clone_cold` now calls `clone()` where the old code deep-copied, a user subclass is silently downgraded to the base class inside the cloned `TimeSeries`.

Everything else checked out. Things I verified rather than assumed:

- The dedented `level is not None` block in `predict` is line-for-line identical to main once `self.ts`→`ts` and `self._cs_df`→`conformity_scores` are normalized. The substitutions are faithful because main already did `self.ts = new_ts` *before* that block, so `self.ts.id_col`/`.uids`/`.ga.n_groups` were already `new_ts`'s.
- The `_cs_df` swap removal is equivalent — the old `finally: if _saved_cs_df is not None` only ran on a path where `self._cs_df` was provably non-None — and no longer leaks on an exception.
- The `scratch = self._with_ts(self.ts._clone_cold())` in `predict`'s transfer path is a real bug fix, not just cleanup: the old `finally: self.ts = _saved_ts_for_cv` rebound `self.ts` to *the same object* `self.preprocess` had just mutated in place, so it restored nothing and `self.ts`'s fitted target-transform state was being clobbered.
- `_with_ts`'s shallow copies don't leak: `fit_models` rebinds `self.models_ = {}` (forecast.py:679) before writing, and `cross_validation` ends with `del self.models_`.
- `as_numpy=False` → `as_numpy=as_numpy` in `fit`'s direct path is inert — `_transform` only consults `as_numpy` inside `if return_X_y:` (core.py:1251) and that call passes `return_X_y=False`.
- The horizons round-trip in `_fit_settings` maps back to the same `(_horizons, max_horizon)` pair through `_validate_horizon_params` for both dense and sparse fits.
- The `Auto*` `_clone` bodies read attributes that exist on coreforecast 0.0.18's scalers (`season_length`, `n_seasons`, `max_season_length`, `max_diffs`), checked against the installed source.
- `_clone_warm` vs. the old hand-rolled `temp_ts` in the fitted-values path: `history_warmup(trim=False)` calls `_initialize_lag_transform_states()`, a superset of the old `_compute_transforms(core_tfms, updates_only=False)`, and `_fit` alone never trimmed either. The one new thing it carries — `horizon_features_` — is inert there, since `_resolve_horizon_features` rejects non-direct mode (forecast.py:340-348) and `forecast_fitted_values` returns early for direct models (forecast.py:1369).
- `_from_ts` matches the construct-then-overwrite it replaces in `load`/`from_cv`/`to_local`; the `{"settings": None, "scores": None}` fallback matches the old `if intervals is not None` guard and now leaves those fields defined rather than unset.

---
_Generated by [Claude Code](https://claude.ai/code)_

nasaul (approved):
LGTM

